PR #13, opened 2015-09-18
redmine:1516 integrating SFI projects reference extraction

marekhorst:
This change is related to project reference extraction feature described in [#1516](https://issue.openaire.research-infrastructures.eu/issues/1516) but it should also fix [#1304](https://issue.openaire.research-infrastructures.eu/issues/1304).

### DIFF
--- a/iis-workflows/iis-workflows-referenceextraction/src/test/java/eu/dnetlib/iis/workflows/referenceextraction/project/WorkflowTest.java
+++ b/iis-workflows/iis-workflows-referenceextraction/src/test/java/eu/dnetlib/iis/workflows/referenceextraction/project/WorkflowTest.java
@@ -1,9 +1,10 @@
 package eu.dnetlib.iis.workflows.referenceextraction.project;
 
-import eu.dnetlib.iis.IntegrationTest;
-import eu.dnetlib.iis.core.AbstractWorkflowTestCase;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+
+import eu.dnetlib.iis.IntegrationTest;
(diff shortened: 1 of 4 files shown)


iis-workflows/iis-workflows-referenceextraction/src/test/java/eu/dnetlib/iis/workflows/referenceextraction/project/WorkflowTest.java
```diff
@@ -19,11 +20,6 @@ public void testMainWorkflow() throws Exception {
 	}
 
     @Test
-	public void testMainSQLiteWorkflow() throws Exception{
```
mkobos: Out of curiosity, why is that test removed?

mkobos:
I finished the review. Looks good to me, but I've got one question (see the comment).

iis-workflows/iis-workflows-referenceextraction/src/test/java/eu/dnetlib/iis/workflows/referenceextraction/project/WorkflowTest.java
marekhorst: It turned out not all tests were failing due to the problem which was patched by Ioannis. 

One test was utilizing predefined sqlite db which become obsolete after minor change in projects model conducted recenty. In order to fix this test we would need to manually recreate db from scratch while in all other tests db was generated on the fly. I just found this test pointless.

mkobos: ok

PR #17, opened 2015-09-26
redmine:1530 fix h2020 project reference extraction issue

marekhorst:


marekhorst:
Do you want to review this?

This is minor change performed by Ioannis and I was about to accept this pull request by myself but maybe we should be more strict and even in such case reassign it to other person.

What do you think?

mkobos:
I think that so far our unwritten policy has been (I guess?) that the changes introduced in partner's code by this partner are his responsibility; we would just check that the integration tests still pass. So I would commit changes like this without doing code reviews. If we agree on that, we can write this rule down in the wiki documentation.

On a related note, I think that you should set the partner (Ioannis in this case) as the author of the commit (`git commit --author="Name Surname <myemail@dot.com>"`).

marekhorst:
Good point, I've just reintroduced commit with proper author name.
Anyway we had lost all original authorship when moving sources to git...

All tests succeeded therefore I am merging branch with master.

mkobos:
Ok.

I added a related wiki entry (section "[1.1 Adding OpenAIRE2020 project's partner code](https://github.com/openaire/iis/wiki/Code-development-workflow)").

PR #20, opened 2015-09-30
Closes #19. Add mechanism to run integration tests workflows on cluster.

### LINKED ISSUE #19: Develop mechanism for integration test workflows run on cluster

madryk:


iis-core/src/test/java/eu/dnetlib/iis/core/AbstractOozieWorkflowTestCase.java
```diff
@@ -0,0 +1,233 @@
...
+	}
+	
+	/**
+	 * Tests workflow located under {@literal workflowLocation} parameter.
+	 * 
+	 * @param workflowLocation - Folder that contains workflow data.
+	 *  	It must contain file {@literal oozie_app/workflow.xml} with workflow definition.
+	 * @param configuration - some additional test pass criteria
+	 *  	(for example timeout - workflow must finish its execution before specified amount of time, 
+	 *   	otherwise test will automatically fail)
```
lukdumi: workflow... before specified amount -> a workflow... within specified time
otherwise test -> otherwise the test

iis-core/src/test/java/eu/dnetlib/iis/core/AbstractOozieWorkflowTestCase.java
```diff
@@ -0,0 +1,233 @@
...
+		
+		
+		String jobId = getJobId();
+		
+		Status jobStatus = waitForJobFinish(jobId, configuration.getTimeoutInSeconds());
+		
+		assertJobStatus(jobId, jobStatus);
+	}
+	
+
```
lukdumi: Can you add the private section comment?

iis-core/src/test/java/eu/dnetlib/iis/core/AbstractOozieWorkflowTestCase.java
```diff
@@ -0,0 +1,233 @@
...
+		
+		String jobId = getJobId();
+		
+		Status jobStatus = waitForJobFinish(jobId, configuration.getTimeoutInSeconds());
+		
+		assertJobStatus(jobId, jobStatus);
+	}
+	
+	
+	private Process runMavenTestWorkflow(String workflowSource) {
```
lukdumi: workflowSource -> workflowPath?

iis-core/src/test/java/eu/dnetlib/iis/core/AbstractOozieWorkflowTestCase.java
```diff
@@ -0,0 +1,233 @@
...
+		log.debug("Setting up OozieClient at {}", getOozieServiceLoc());
+		oozieClient = new OozieClient(getOozieServiceLoc());
+	}
+	
+	/**
+	 * Tests workflow located under {@literal workflowLocation} parameter.
+	 * Internally uses {@link #testWorkflow(String, OozieWorkflowTestConfiguration)}
+	 * with default {@link OozieWorkflowTestConfiguration}
+	 */
+	protected void testWorkflow(String workflowLocation) {
```
lukdumi: I wonder if we shouldn't name this param 'workflowPath'? 'Location' is a very general term and here we just want a path to the workflow dir.

iis-core/src/test/java/eu/dnetlib/iis/core/AbstractOozieWorkflowTestCase.java
```diff
@@ -0,0 +1,233 @@
...
+		}
+		Pattern pattern = Pattern.compile("^job: (\\S*)$", Pattern.MULTILINE);
+		Matcher matcher = pattern.matcher(jobId);
+		matcher.find();
+		jobId = matcher.group(1);
+		
+		return jobId;
+	}
+	
+	private Status waitForJobFinish(String jobId, long timeoutInSeconds) {
```
lukdumi: there's something wrong with formatting here in this method.

iis-core/src/test/java/eu/dnetlib/iis/core/AbstractOozieWorkflowTestCase.java
```diff
@@ -0,0 +1,233 @@
...
+		}
+    	Assert.fail("Execution of job " + jobId + " exceeded waiting time limit");
+    	
+    	return null;
+	}
+	
+	private void assertJobStatus(String jobId, Status status) {
+		if (status == Status.FAILED || status == Status.KILLED || status == Status.SUSPENDED) {
+			
+			try {
```
lukdumi: What about moving this 'try catch' to a separate private method. You could also use it in waitForJobFinish.

iis-core/src/test/java/eu/dnetlib/iis/core/IntegrationTestPropertiesReader.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+public class IntegrationTestPropertiesReader {
+
+	private final static String DEFAULT_PROPERTIES_CLASSPATH = "/integration-test-default.properties";
+	
+	private final static String USER_PROPERTIES_PATH = System.getProperty("user.home") + "/.iis/integration-test.properties";
+	
+	private Properties integrationTestProperties = new Properties();
+	
+
```
lukdumi: public/private 'saos-like' sections

iis-core/src/test/java/eu/dnetlib/iis/core/IntegrationTestPropertiesReader.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.InputStream;
+import java.util.Properties;
+
+import org.apache.commons.io.IOUtils;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * Reader of properties for integration tests.<br/>
+ * It uses file {@literal classpath:integration-test-default.properties} as source of properties
```
lukdumi: ...as a source of...

iis-core/src/test/java/eu/dnetlib/iis/core/IntegrationTestPropertiesReader.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Properties;
+
+import org.apache.commons.io.IOUtils;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * Reader of properties for integration tests.<br/>
+ * It uses file {@literal classpath:integration-test-default.properties} as source of properties 
+ * and allow to override them in ${user.home}/.iis/integration-test.properties
```
lukdumi: ...allows for overriding them...

iis-core/src/test/java/eu/dnetlib/iis/core/IntegrationTestPropertiesReader.java
```diff
@@ -0,0 +1,93 @@
...
+	    	throw new RuntimeException(e);
+	    } finally {
+	    	IOUtils.closeQuietly(inputStream);
+	    }
+	    
+	    return defaultProperties;
+	}
+	
+	private Properties readUserProperties() {
+		Properties userOverridenProperties = new Properties();
```
lukdumi: userProperies?

iis-workflows/pom.xml
```diff
@@ -56,20 +56,15 @@
         <oozie.package.file.name>oozie-package</oozie.package.file.name>
         <!-- notice: sandboxName is generated based on workflow.source.dir property -->
         <workflow.source.dir>src/test/resources/define/path/pointing/to/directory/holding/oozie_app</workflow.source.dir>
```
lukdumi: I don't think it makes any sense to provide this artficial value for the workflow.source.dir. The plugin that needs it should fail if the value is not provided in the command line - not defining it should make the trick, shouldn't it?

iis-core/src/test/java/eu/dnetlib/iis/core/IntegrationTestPropertiesReader.java
madryk: :+1:

iis-core/src/test/java/eu/dnetlib/iis/core/IntegrationTestPropertiesReader.java
madryk: ok

iis-core/src/test/java/eu/dnetlib/iis/core/IntegrationTestPropertiesReader.java
madryk: :+1:

iis-core/src/test/java/eu/dnetlib/iis/core/AbstractOozieWorkflowTestCase.java
madryk: ok, i will correct it.

iis-core/src/test/java/eu/dnetlib/iis/core/AbstractOozieWorkflowTestCase.java
madryk: ok

iis-core/src/test/java/eu/dnetlib/iis/core/AbstractOozieWorkflowTestCase.java
madryk: Ok. I Agree.

iis-core/src/test/java/eu/dnetlib/iis/core/AbstractOozieWorkflowTestCase.java
madryk: :+1:

iis-core/src/test/java/eu/dnetlib/iis/core/AbstractOozieWorkflowTestCase.java
madryk: :+1:

iis-workflows/pom.xml
madryk: Not exactly. Maven is flexible with properties. If he finds something that look like property placeholder (eg. `${workflow.source.dir}`) but property doesn't exist, then it would leave it as is and continue build.
In our case it would result in priming that will lead to success, but build will fail in next stages. Error will say something about circular reference in property definitions which don't point us directly to source problem.
When `workflow.source.dir` is provided in pom, then build will fail while priming. Giving us message that `src/test/resources/define/path/pointing/to/directory/holding/oozie_app` was not in classpath. So cause of build fail is easier to find.

lukdumi: OK, so let's leave it as it is.

iis-core/src/test/java/eu/dnetlib/iis/core/AbstractOozieWorkflowTestCase.java
madryk: ok

iis-core/src/test/java/eu/dnetlib/iis/core/IntegrationTestPropertiesReader.java
madryk: ok

PR #45, opened 2015-10-10
Closes #8: add license

### LINKED ISSUE #8: Add license
We should add information about Apache 2.0 license to the code. This requires

1. adding two files: `NOTICE` and `LICENSE` in the project's root directory,
2. adding a short license header to all source files (although I'm not sure if it's really necessary, since in [1] they say that it "should" be done, which is in contrast to description in [1] of adding `NOTICE` file which "must" be done)
3. updating README file with an information about the license of the project.

Additional material - information about applying Apache license in your project on the web:

[1] http://www.apache.org/dev/apply-license.html#new
[2] http://blog.maestropublishing.com/2009/11/19/how-to-apply-the-apache-2-0-license-to-your-project/
[3] take a look how it was done in the [avroknife project](https://github.com/CeON/avroknife).

mkobos:
Please review. I guess nothing interesting here really.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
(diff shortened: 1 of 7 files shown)


marekhorst:
Looks fine. Shouldn't we do the same for primer?

mkobos:
Good idea

Issue #23, opened 2015-09-30
Remove MiniOozie from project


madryk:
We will not be using it anymore.
Remove any code that is related to it.

PR #46, opened 2015-10-11
#23: Remove mini oozie

madryk:


lukdumi:
Wow, I didn't expect that we would get rid of so many lines ;-)

Issue #38, opened 2015-10-08
Improve running oozie packages from maven


madryk:
Implement according to conclusions made in #26 
- remove `application-default.properties`
- running single test: `mvn integration-test -Dtest=eu.dnetlib...TestClass`
- remove `deploy-local`, `run-local` profiles (all workflows are deployed by ssh)
- additional parameter for deploying workflows `-DconnectionProperties=/path/to/cluster/properties`
- integration tests are configured to use cluster configured in `~/.iis/integration-tests.properties`, but can be overriden by `-DconnectionProperties` property: `mvn integration-test -DconnectionProperties=/path/to/cluster/properties`

PR #47, opened 2015-10-12
Improve integration tests

madryk:
#38


iis-core/src/test/java/eu/dnetlib/iis/core/IntegrationTestPropertiesReader.java
```diff
@@ -38,6 +50,8 @@ public IntegrationTestPropertiesReader() {
 	    
 	    integrationTestProperties.putAll(defaultProperties);
 	    integrationTestProperties.putAll(userProperties);
+	    
+	    createTempPropertiesFile();
```
lukdumi: This constructor doesn't look elegant. It implicitly reads properties, creates files. Apart from that - the class is not focused on one thing. Let's talk about it at our meeting.

iis-core/src/test/java/eu/dnetlib/iis/core/IntegrationTestPropertiesReader.java
```diff
@@ -97,4 +127,27 @@ private Properties readUserProperties() {
 		return userProperties;
 	}
 	
+	private String resolveUserPropertiesFilePath() {
+		String userPropertiesPath = System.getProperty("connectionProperties");
```
lukdumi: I wonder if the "connectionProperties" name is not too generic. iisIntTestConnectionProperties?

iis-core/src/test/java/eu/dnetlib/iis/core/IntegrationTestPropertiesReader.java
```diff
@@ -97,4 +127,27 @@ private Properties readUserProperties() {
...
+		if (userPropertiesPath == null) {
+			return DEFAULT_USER_PROPERTIES_PATH;
+		}
+		return userPropertiesPath;
+	}
+	
+	private void createTempPropertiesFile() {
```
lukdumi: I think it should be in a separate service method that takes Properties and file path as parameters.

iis-workflows/pom.xml
```diff
@@ -686,47 +606,49 @@
...
 							</goals>
 							<configuration>
-								<skip>false</skip>
-								<forkMode>always</forkMode>
+								<redirectTestOutputToFile>true</redirectTestOutputToFile>
 								<argLine>-Xmx1024m</argLine>
 								<systemPropertiesVariables>
 									<hadoop.log.dir>/tmp</hadoop.log.dir>
 									<oozie.data.dir>${oozie.data.dir}</oozie.data.dir>
+									<connectionProperties>${connectionProperties}</connectionProperties>
```
lukdumi: Information that if the connectionProperties is null then default connection properties (with location of it) will be used, would be useful.

madryk:
I made all corrections.

madryk:
I've applied all changes

Issue #37, opened 2015-10-08
introduce exporter module serializing actions to JSON as described in redmine:1522


marekhorst:
More details in:

https://issue.openaire.research-infrastructures.eu/issues/1522#note-15

marekhorst:
Newly introduced functionality is activated when `output_remote_location` parameter pointing to remote HDFS location is provided. HBase export mode is activated when parameter was not set.

PR #57, opened 2015-10-16
marekhorst #37 exporter module serializing actions to json

marekhorst:
This feature is related to new CDH5->CDH4 compatible exporter module writing actionsets as sequence files with JSON records instead of exporting them directly as protobufs to HBase.

More details can be found in #37 issue.

Please review.

### DIFF
--- a/iis-core/src/main/scripts/oozie2dot/examples/example_workflow/a_subworkflow/workflow.xml
+++ b/iis-core/src/main/scripts/oozie2dot/examples/example_workflow/a_subworkflow/workflow.xml
@@ -792,8 +792,8 @@
 			<arg>eu.dnetlib.iis.workflows.export.actionmanager.entity.dataset.DatasetExporterProcess</arg>
 			<arg>-Iinput=${workingDir}/identifier/datasets</arg>
 			
-			<arg>-Pimport.datacite.mdstore.service.location=${import_mdstore_service_location}</arg>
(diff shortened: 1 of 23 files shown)


mkobos:
I think that we should stick to the convention that the pull request should be named the same as the commit corresponding to this request and the commit itself should have a form of "Closes #$ISSUE_NUMBER".

mkobos:
I'm done with the review, please take a look.

mkobos:
One more thing: we should have some tests covering this functionality.

marekhorst:
Regarding the tests: I am all for it. This will take some time, so I will suggest creating separate issue for this task.

marekhorst:
I've just pushed bunch of fixes requested in comments.

mkobos:
> Regarding the tests: I am all for it. This will take some time, so I will suggest creating separate issue for this task.

Agree. Taking into consideration that we've got a lot of close deadlines upon us, this is the only option. I think that in the future, the tests should be a standard part of pull requests.

mkobos:
I'm done with answering, please take a look.

marekhorst:
Sure. I've created #73 ticket to track test related task.

marekhorst:
I've just committed lighter version of `run()` method.

mkobos:
Still, at ~50 lines not exactly a short method, but more readable. 

Looks good to me.

Issue #52, opened 2015-10-14
introduce Faults support in EuropePMC importer module


marekhorst:
Originally reported in:
https://issue.openaire.research-infrastructures.eu/issues/1415

In this task we should introduce `Fault` datastore creation when any exception occurs during PMC metadata extraction. This was already done for PDF metadata extraction (cermine).

This is improved approach of simple logging which is not visible enough and cannot be treated as a subject for further processing and in-depth analysis.

PR #58, opened 2015-10-16
marekhorst 52 fault support in epmc importer

marekhorst:
Related to #52 issue. Please review.


iis-common/src/main/java/eu/dnetlib/iis/common/fault/FaultUtils.java
```diff
@@ -0,0 +1,61 @@
...
+		StringWriter strWriter = new StringWriter();
+		PrintWriter pw = new PrintWriter(strWriter);
+		throwable.printStackTrace(pw);
+		pw.close();
+		faultBuilder.setStackTrace(strWriter.toString());
+		if (throwable.getCause()!=null) {
+			faultBuilder.setCauses(appendThrowableToCauses(
+					throwable.getCause(), new ArrayList<Cause>()));
+		}
+		if (auditSupplementaryData!=null && !auditSupplementaryData.isEmpty()) {
```
mkobos: Do we have to check whether the `auditSupplementaryData!=null`? Can't we just disallow it by setting appropriate assertion at the beginning of the method, e.g. using Guava's `com.google.common.base.Preconditions.checkNotNull()`?

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
```diff
@@ -46,15 +85,29 @@ protected void map(AvroKey<DocumentText> key, NullWritable value, Context contex
...
+				mos.write(namedOutputMeta, 
+						new AvroKey<ExtractedDocumentMetadata>(output.build()));
         	 } catch (ParserConfigurationException e) {
-             	throw new IOException("Text extraction failed for id " + nlm.getId() + 
-             			" and text: " + nlm.getText(), e);
+        		 Map<CharSequence, CharSequence> auditSupplementaryData = new HashMap<CharSequence, CharSequence>();
+        			auditSupplementaryData.put(FAULT_TEXT, nlm.getText());
+        		 mos.write(namedOutputFault, 
+        				 new AvroKey<Fault>(FaultUtils.exceptionToFault(nlm.getId(), e, auditSupplementaryData)));
         	 } catch (SAXException e) {
```
mkobos: Indentation of the following section is too deep.

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
```diff
@@ -46,15 +85,29 @@ protected void map(AvroKey<DocumentText> key, NullWritable value, Context contex
...
 						pmcXmlHandler);
-                context.write(new AvroKey<ExtractedDocumentMetadata>(output.build()), 
-                		NullWritable.get());
+				mos.write(namedOutputMeta, 
+						new AvroKey<ExtractedDocumentMetadata>(output.build()));
         	 } catch (ParserConfigurationException e) {
-             	throw new IOException("Text extraction failed for id " + nlm.getId() + 
-             			" and text: " + nlm.getText(), e);
+        		 Map<CharSequence, CharSequence> auditSupplementaryData = new HashMap<CharSequence, CharSequence>();
+        			auditSupplementaryData.put(FAULT_TEXT, nlm.getText());
```
mkobos: Is there a limited set of keywords that can be used as keys of the `auditSupplementaryData` map? Answer to this question should clearly given in the `Fault` Avro schema. Why is `FAULT_TEXT=="text"` used as key here?

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
```diff
@@ -46,15 +85,29 @@ protected void map(AvroKey<DocumentText> key, NullWritable value, Context contex
...
+        		 mos.write(namedOutputFault, 
+        				 new AvroKey<Fault>(FaultUtils.exceptionToFault(nlm.getId(), e, auditSupplementaryData)));
         	 } catch (SAXException e) {
- 				throw new IOException("Text extraction failed for id " + nlm.getId() + 
-             			" and text: " + nlm.getText(), e);
+				Map<CharSequence, CharSequence> auditSupplementaryData = new HashMap<CharSequence, CharSequence>();
```
mkobos: The section starting here is the same one as in the previous `catch` block. The code should be reused.

iis-workflows/iis-workflows-ingest-pmc/src/main/resources/eu/dnetlib/iis/workflows/ingest/pmc/metadata/oozie_app/workflow.xml
```diff
@@ -33,18 +43,13 @@
             <name-node>${nameNode}</name-node>
  			<!-- The data generated by this node is deleted in this section -->
 			<prepare>
-				<delete path="${nameNode}${output}" />
+				<delete path="${nameNode}${output_root}" />
 			</prepare>
             <configuration>
             	<property>
 					<name>mapreduce.inputformat.class</name>
```
mkobos: Mixed indentation.

iis-workflows/iis-workflows-top/src/main/resources/eu/dnetlib/iis/workflows/top/common/import/job.properties
```diff
@@ -12,21 +12,18 @@ project_concepts_context_id=fet-fp7
 database_service_location=http://beta.services.openaire.eu:8280/is/services/database
```
mkobos: We should remove hardcoded addresses and identifiers like this from the code and place them in some external configuration.

iis-workflows/iis-workflows-top/src/main/resources/eu/dnetlib/iis/workflows/top/common/import/oozie_app/workflow.xml
```diff
@@ -710,16 +718,26 @@
         <java>
 			<prepare>
```
mkobos: Strange indentation. This whole file should be reindented to make the indentation uniform.

mkobos:
I think that we should stick to the convention that the pull request should be named the same as the commit corresponding to this request and the commit itself should have a form of "Closes #$ISSUE_NUMBER".

mkobos:
I'm done with the review. Please take a look.

iis-common/src/main/java/eu/dnetlib/iis/common/fault/FaultUtils.java
marekhorst: This is optional parameter, not required. The main reason for this `if` statement is to leave `faultBuilder` untouched if `auditSupplementaryData` doesn't convey any data.

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
marekhorst: Do we know all possible scenarios of `auditSupplementaryData` usage? I don't know this is why I didn't mention about possible values. So far I've been using `text` and `processing_time` each one in different context and processing module. We could describe it e.g. in schema but this will beg for complex explanation and implicitly require schema comment updates whenever new audit data type is used. I am not sure schema is the right place for this.

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
marekhorst: I just fixed this in local branch and will push it in a while.

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
marekhorst: Good point. I just fixed this in local branch and will push it in a while.

iis-workflows/iis-workflows-ingest-pmc/src/main/resources/eu/dnetlib/iis/workflows/ingest/pmc/metadata/oozie_app/workflow.xml
marekhorst: I just fixed this in local branch and will push it in a while.

iis-workflows/iis-workflows-top/src/main/resources/eu/dnetlib/iis/workflows/top/common/import/job.properties
marekhorst: I will fix this by putting `job.properties` to `.gitignore`, already reported in different pull request.

iis-workflows/iis-workflows-top/src/main/resources/eu/dnetlib/iis/workflows/top/common/import/oozie_app/workflow.xml
marekhorst: Fixed that. Although fixing this is least favorite task of mine... I hope it won't make much pain when merging changes coming from different branches because indention changes will be picked up.

marekhorst:
I've done answering to comments and pushed set of fixes.

iis-common/src/main/java/eu/dnetlib/iis/common/fault/FaultUtils.java
mkobos: Why do we strive to leave `faultBuilder` untouched?

iis-workflows/iis-workflows-top/src/main/resources/eu/dnetlib/iis/workflows/top/common/import/job.properties
mkobos: Ok

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
mkobos: To me, the solution with updating comments in schema seems fine. Currently, the knowledge about possible keys and their meaning is distributed in codebase which I don't like. What do you think?

mkobos:
I'm done with answering, please take a look.

iis-common/src/main/java/eu/dnetlib/iis/common/fault/FaultUtils.java
marekhorst: I don't want to set empty map if it does not contain any values. I am OK with having `null` here rather than having empty map instance serialized by avro.

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
marekhorst: I am also fine with this solution as soon as everyone remembers about taking care of updating this comment whenever set of supported keys changes. How can we guarantee this? 

Besides, currently only theoretically, If we want to keep posted all the people depending on  `iis-schema` module lib in their dependencies this will also require us to release this module only due to new comment was added.

marekhorst:
I am done with my answers.

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
mkobos: > I am also fine with this solution as soon as everyone remembers about taking care of updating this comment whenever set of supported keys changes. How can we guarantee this? 

I assume that when someone introduces a new supported key, he works with the schema anyway. In this case he would take a look at the schema and see that adding comment is required.

> Besides, currently only theoretically, If we want to keep posted all the people depending on `iis-schema` module lib in their dependencies this will also require us to release this module only due to new comment was added. 

True, this would be a small change but a change in schema nevertheless. A direct counterpart of it would be a modification of values in Avro's `Enum` and this would mean actually changing the code of schema. Here it would be best to have a possibility of defining keys as enums but Avro doesn't allow this, so we have to stick to comments.

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
mkobos: ok

iis-workflows/iis-workflows-ingest-pmc/src/main/resources/eu/dnetlib/iis/workflows/ingest/pmc/metadata/oozie_app/workflow.xml
mkobos: ok

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
mkobos: ok

iis-common/src/main/java/eu/dnetlib/iis/common/fault/FaultUtils.java
mkobos: ok

mkobos:
I'm done with my answers, please take a look.

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
marekhorst: I could go with the list of supported keys but would it be meaningful without mentioning in which context they do appear? 

Mentioning about context will lead to comments related to other modules such as:
* `processing_time` - provided by pdf metadataextraction
* `url` - used by pdf metadataextraction to point to the source url of the content
* `text` -  used by pmc ingester, conveys pmc source xml file

what doesn't look good in such generic module as `iis-schemas`.

Besides, `auditSupplementaryData` is self explanatory - this is an auxiliary data that can be helpful in debugging fault, but it is not an essential part of `Fault` model. 

Maybe fault dedicated wiki page will be a better place for such details?

marekhorst:
I am done with my answers. 

I guess we could use `Assignee` field to indicate who should step in. What do you think?

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
mkobos: I get your point. 

On the other hand, I don't like very flexible structures like this where you can throw almost everything into. I think that what you're doing here is trying to get a functionality of class inheritance from Object Oriented languages, i.e. make the structure open for extending. In OO in case of metadataextraction, we would have `MetadataExtractionFault` class that would inherit from `Fault` and add two additional fields: `processing_time` and `url` while in case of PMC ingester we would have `PMCIngestionFault` class that would inherit from `Fault` and add field `text`.

A solution that would provide (poor man's) inheritance might look like this:

    record BasicFaultInfo {
       string inputObjectId;
       long timestamp;
       string code;
       // (...)
    }

    record Fault {
       BasicFaultInfo core;
    }

    record MetadataExtractionFault {
       BasicFaultInfo core;
       int processing_time;
       string url;
    }

    record PMCIngestionFault {
       BasicFaultInfo core;       
       string text;
    }

`BasicFaultInfo` would be placed in a common `audit` namespace (just where `Fault` is now) while `MetadataExtractionFault` and `PMCIngestionFault` would be placed in namespaces corresponding to metadataextraction and pmcingestion, respectively.

A disadvantage of this approach is that it is more complex, however the advantages are:

- auxiliary fields are explicitly defined
- the auxiliary fields might be records
- auxiliary fields have proper types defined

mkobos:
I'm done with my answers.

> I guess we could use Assignee field to indicate who should step in. What do you think?

Good idea. Let's try.

iis-workflows/iis-workflows-ingest-pmc/src/main/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/MetadataImporter.java
marekhorst: This is very good analogy. If this was proper inheritance then we could use common tool for handling `Fault` objects. Here we would have different objects for each scenario (only with shared subset of fields) so if we finally prepare backend fault processing utility we would need to tailor it specifically for each fault sink. 

Now I tend towards postponing the agreements related to the final `Fault` model until we have clear vision of what we are going to do with the faults in the end. Up until now the only usage scenario was to group & filter faults in hive and calculate statistics.

mkobos: > This is very good analogy. If this was proper inheritance then we could use common tool for handling Fault objects. Here we would have different objects for each scenario (only with shared subset of fields) so if we finally prepare backend fault processing utility we would need to tailor it specifically for each fault sink. 

Even in this case the generic processing utility should work properly - it would access only the `core` field which would be common for all `Fault`s.

marekhorst: This is surprising, are you sure this implicit inheritance will work? Does it work by design or by accident (the way avro currently serializes objects)? What if the fields ordering is changed in a way `BasicFaultInfo` is not the first field in record?

If it is safe to relay on this feature I guess we could agree on introducing this extendable `Fault` model. This will be both: explicitly defined and flexible approach.

mkobos: Ok, It's not that easy as I thought. According to specification of [Avro schema resolution][1], in order to make a generic code read `MetadataExtractionFault` and `PMCIngestionFault` records as if each of them was a `Fault` record (polymorphism-like behaviour), both of these former records would have to be named `Fault` and come from the same namespace as `Fault`. 

I think that our code that generates java classes from Avro schemas wouldn't like this setting. Apart from that, it might be confusing to have different structures with the same name.

BTW, field ordering doesn't matter - according to [Avro schema resolution][1] "the ordering of fields may be different: fields are matched by name."

[1]: http://avro.apache.org/docs/1.7.7/spec.html#Schema+Resolution

marekhorst: OK, so I guess we are back to the idea of creating dictionary for supplementary data.

What do you think about putting link to wiki page in `Fault` schema comment? This way we will be more flexible and still have this information close to the code.

mkobos: I think it would be even better to keep this schema-related information directly in our code base. I'm thinking about putting comment on the output port producing `Fault` structure in pmcingestion and metadataextraction modules that says what `supplementaryData` entries are produced. To be more precise, this comment should probably go into `parameters/property/description` field of Oozie XML defining subworkflows corresponding to pmcingestion and metadataextraction modules.

I would also extend the comment describing `supplementaryData` field in the `Fault` structure to something like this: "Other supplementary data related to specific type of fault. See in-code documentation (in Oozie XML) of modules that use this structure for information what exactly they put in here."

What do you think?

marekhorst: Let it be. Just pushed the commit.

mkobos:
Looks good to me.

PR #64, opened 2015-10-20
Closes #62. Connect to oozie through ssh in integration tests

### LINKED ISSUE #62: Check oozie status in integration tests via ssh
instead of using OozieClient

madryk:


iis-core/src/test/java/eu/dnetlib/iis/core/SshOozieClient.java
```diff
@@ -0,0 +1,133 @@
...
+		if (sshClient != null) {
+			sshClient.close();
+			sshClient = null;
+		}
+	}
+	
+	/**
+	 * Returns job status with provided id
+	 */
+	public Status getJobStatus(String jobId) throws IOException {
```
lukdumi: I wonder if we should not change the IOException to a runtime one here.

iis-core/src/test/java/eu/dnetlib/iis/core/SshExecUtils.java
```diff
@@ -0,0 +1,81 @@
...
+	 * 
+	 * @param sshClient - client used to execute command
+	 * @param command - command that will be executed on remote machine
+	 * @param throwExceptionOnCommandError - if true then any error in executing
+	 * 		command on remote machine will cause this method to throw exception
+	 * @return command execution results
+	 * @throws IOException - in case of problems with connection or data transmission
+	 * 		between local and remote machines
+	 */
+	public static Command sshExec(SSHClient sshClient, String command, boolean throwExceptionOnCommandError) throws IOException {
```
lukdumi: Shouldn't we rethrow the exception here as runtime one?

iis-core/src/test/java/eu/dnetlib/iis/core/SshOozieClient.java
```diff
@@ -0,0 +1,133 @@
...
+	
+	/**
+	 * Opens ssh connection.<br/>
+	 * Method must be executed before any attempt to read from remote host.
+	 */
+	public void openConnection() throws IOException {
+		if (sshClient != null) {
+			log.warn("Attempt to open new connection when the old one is still opened. Will use old connection.");
+			return;
+		}
```
lukdumi: In my opinion it would be better to throw and exception if one wants to connect second time (without first closing the connection).

iis-core/src/test/java/eu/dnetlib/iis/core/SshBasedHdfsFileFetcher.java
```diff
@@ -20,8 +16,6 @@
```
lukdumi: I would shorten the name to SshHdfsFileFetcher.

lukdumi:
It starts to look like a nice piece of software. I am thinking about moving it to a separate package in src/main and writing some tests for it. Let's talk about it at our meeting today.

Issue #76, opened 2015-10-26
integrate initial sentiment analysis implementation with IIS


marekhorst:
Specification described in:
https://issue.openaire.research-infrastructures.eu/issues/1443

Initially sentiment analysis mock should be integrated.

PR #78, opened 2015-10-26
Marekhorst #1443 citation sentiment analysis integration

marekhorst:
This pull request is related to extensively discussed sentiment analysis integration from #76.

Please review.


mkobos:
I looked through the changes. I reported mainly comments about inconsistent indentation, so I guess I haven't looked hard enough (but this pull request is huge!)

marekhorst:
I've just pushed commit with all the requested changes.

marekhorst:
@mkobos can you scan briefly my last commit so I could finally merge it with master? 

Don't be scared, there are lots of changes but most of them are indention related (all `workflow.xml` and `*.avdl` files) so I guess you could skip them.

mkobos:
Looks good to me.

iis-workflows/iis-workflows-sentimentanalysis/src/main/java/eu/dnetlib/iis/workflows/sentimentanalysis/SentimentAnalyzer.java
```diff
@@ -0,0 +1,192 @@
...
+     * @param publicationId
+     * @param references
+     * @param sentiments
+     * @return citation sentiments converted into
+     *         {@link ReferenceWithSentimentLabels} objects
+     */
+    protected static List<ReferenceWithSentimentLabels> convertCitationSentiment(CharSequence publicationId,
+            List<ReferenceMetadata> references, List<CitationSentiment> sentiments) {
+        Preconditions.checkArgument(references.size() != sentiments.size(), "number of sentiments: " + sentiments.size()
+                + " is not equal the number of input references: " + references.size());
```
mkobos: Please use `Preconditions.checkArgument(references.size() != sentiments.size(), "number of sentiments: %s is not equal the number of input references: %s", sentiments.size(), references.size())`, otherwise the string will be constructed every time the method is called. I had a problem like this before and it ended in a huge performance hit.

marekhorst: Done. This could be the problem indeed: string is constructed at each precondition's check not only when check fails.

PR #99, opened 2015-11-06
Closes #98: Exclude tar.gz dependencies from uber jars created by shade plugin [cdh5]

### LINKED ISSUE #98: Exclude tar.gz dependencies from uber jars created by shade plugin [cdh5]
You can recreate this issue by executing:
`mvn clean package -Poozie-package -Dworkflow.source.dir=eu/dnetlib/iis/workflows/top/primary/main`
in `iis-workflows/iis-workflows-top`.

All coansys dependencies are fat `tar.gz` packages including oozie workflows along with libs.

Primer plugin is capable of maintaining such dependencies by extracting them and attaching to application wherever the following line in `import.txt` files was defined:
```
coansys coansys pl.edu.icm.coansys:citations-core-workflow
```
All jars from those `tar.gz` packages are required by coansys submodules only. Neither `tar.gz` files nor jars included in those packages have to become part of uber-jar.

lukdumi:


lukdumi:
Marek, I have assigned this PR to you, as you are the person who has reported this bug. 
Now the shade plugin adds only jars to uber jars. This is the simplest solution and maven does not complain now. The only problem I can see now is that the uber files are somewhat big but I do not think we can do something with it.

marekhorst:
Sorry for late reply. Looks pretty straightforward to me. Lets bring those changes to `cdh5` branch.

PR #107, opened 2015-11-12
Closes #93: add job.properties to .gitignore

### LINKED ISSUE #93: add job.properties to .gitignore
To make project clean we shouldn't keep `job.properties` files in git therefore we should ignore them all e.g. by putting `**/job.properties` in `.gitignore`. We would need to remove all files from git control first.

This will make workflows manual deployment slightly more difficult but we can find solutions for this problem, e.g. by keeping `job.properties` in mirrored directory structure somewhere aside the git folder and creating symbolic links within git folders. This will prevent us from accidental `job.properties` removal when purging whole git directory.

marekhorst:
Can you review this pull request? This one is quite simple.


mkobos:
> This one is quite simple.

Sounded enticing :-) This commit was really simple. However, how will all these ignored `job.properties` files be provided to the workflows? In #93 you mention some solution with symbolic links and this sounds puzzling and not system-independent.

marekhorst:
> However, how will all these ignored job.properties files be provided to the workflows? In #93 you mention some solution with symbolic links and this sounds puzzling and not system-independent.

Those `job.properties` files were required only for manual jobs execution, in fact the ones from `iis-workflows-top` were mostly in use (at least by me) the other were used occasionally.

Each developer can find most suitable approach, the one described in #93 relies on symbolic links to prevent permanent `job.properties` removal when e.g. purging cloned repository from local filesystem (they will permanently disappear because they are not tracked by git). I prefer to store `job.properties` outside and link to those files in IIS project structure.

mkobos:
> Each developer can find most suitable approach, the one described in #93 relies on symbolic links to prevent permanent job.properties removal when e.g. purging cloned repository from local filesystem (they will permanently disappear because they are not tracked by git). I prefer to store job.properties outside and link to those files in IIS project structure.

Interesting topic. We can talk about it if an opportunity arises.

PR #133, opened 2015-11-25
Closes #122: Create an independent project for spark-utils

### LINKED ISSUE #122: Create an independent project for spark-utils

lukdumi:


lukdumi:
You can find the spark utils project here:
https://github.com/CeON/spark-utils

pom.xml
```diff
@@ -377,6 +378,14 @@
                 <version>1.0.0</version>
             </dependency>
             
+            <dependency>
+                <groupId>pl.edu.icm.spark-utils</groupId>
+                <artifactId>spark-utils</artifactId>
+                <version>0.0.1-SNAPSHOT</version>
+            </dependency>
```
madryk: spark-utils define spark-core_2.10 and spark-sql_2.10 dependencies.
So now we will have spark libraries in runtime classpath and they will be packaged in production workflows.
Two solutions come to my mind. Don't know witch is better.
1) exclude spark on above dependency
2) mark spark scope as provided in spark-utils

lukdumi: Yes, I thought about it a little... I am not sure if it's a problem now, because we have explicitely defined the scope of these depenedencies as provided (in iis) and the project scope will take precendence.

madryk: Oh. You're right.

PR #138, opened 2015-11-26
Closes #137: upgrade cermine dependency to 1.8-SNAPSHOT

### LINKED ISSUE #137: upgrade cermine dependency to 1.8-SNAPSHOT
We will keep up with the most recent version and bring `bcprov-jdk15on` dependency instead of `jdk14`.

marekhorst:
and fix integration test


iis-workflows/iis-workflows-metadataextraction/src/test/resources/eu/dnetlib/iis/workflows/metadataextraction/sampledataproducer/data/extr_meta.json
```diff
@@ -499,5 +503,5 @@
...
+  "text": ""
```
marekhorst: @dtkaczyk it seem cermine was unable to extract text after upgrading dependency to 1.8-SNAPSHOT. Was it expected?

PR #147, opened 2015-12-03
Closes #144: fix out of memory errors in basic collapser

### LINKED ISSUE #144: fix out of memory errors in basic collapser
Apparently default amount on allocated memory (200MB) turns out to be not enough for basic collapser causing:

```
Error: java.lang.OutOfMemoryError: Java heap space
```

when collapsing extracted metadata records. Size of each record increased after adding `text` field after integrating sentiment analysis in #76.

We should increase maximum memory to be allocated it in the same way it was done in origin collapser by setting `mapred.child.java.opts` property to `-Xmx2048M` by default. It should be also possible to override this value.

marekhorst:
Main changes are related to newly introduced `mapred.child.java.opts` parameter. All the others are caused by indentation fix.


iis-workflows/iis-workflows-collapsers/src/main/resources/eu/dnetlib/iis/workflows/collapsers/origins_collapser/oozie_app/workflow.xml
```diff
@@ -18,60 +18,64 @@
...
-                    <name>mapred.reduce.tasks</name>
-                    <value>36</value>
-                </property>
-                <property>
-                    <name>mapred.child.java.opts</name>
-                    <value>-Xmx2048M</value>
-                </property>
+				<property>
+					<name>mapred.reduce.tasks</name>
+					<value>36</value>
```
mkobos: Magic number?

iis-workflows/iis-workflows-collapsers/src/main/resources/eu/dnetlib/iis/workflows/collapsers/basic_collapser/oozie_app/workflow.xml
```diff
@@ -14,49 +14,59 @@
...
+			<value>$ALL$</value>
 			<description>comma-separated list of significant field names used in within origin collapser;
-                         @see eu.dnetlib.iis.workflows.collapsers.SignificantFieldsCollapser</description>
+				@see eu.dnetlib.iis.workflows.collapsers.SignificantFieldsCollapser
+			</description>
+		</property>
+		<property>
+			<name>mapred_child_java_opts</name>
+			<value>-Xmx2048M</value>
+			<description>java-opts, e.g. maximum heap size for oozie</description>
```
mkobos: Instead of "maximum heap size for oozie" it should be rather something like "maximum heap size for a single JVM running MapReduce".

mkobos:
The changes are simple, but due to changes in indentation it is difficult to find where they really are. I think it will be better to extract changes to the indentation into a separate commit within a pull request next time.

iis-workflows/iis-workflows-collapsers/src/main/resources/eu/dnetlib/iis/workflows/collapsers/basic_collapser/oozie_app/workflow.xml
marekhorst: Mental shortcut went too far :)

iis-workflows/iis-workflows-collapsers/src/main/resources/eu/dnetlib/iis/workflows/collapsers/origins_collapser/oozie_app/workflow.xml
marekhorst: Yeah, that's interesting. 

I went through svn history and found this magic number appearing in `r29647` after Dominika's `memory-related properties set` commit. It was on 2014-07-29 so it seems it was adjusted to current datanodes number. 

We could try not to hardcode it but to use placeholder and set it somewhere in top level workflow. I would also recommend creating dedicated ticket for this.

marekhorst:
> The changes are simple, but due to changes in indentation it is difficult to find where they really are. I think it will be better to extract changes to the indentation into a separate commit within a pull request next time.

I agree, we should write it in a stone ;)

marekhorst:
I've just committed updates, please take a look.

mkobos:
LGTM

PR #156, opened 2015-12-18
Closes #155: adjust metadataextraction logging level

### LINKED ISSUE #155: adjust metadataextraction logging level to show identifiers of processed documents
Currently we cannot easily find the identifiers of documents which were processed by cermine. 

This makes debugging process rather difficult and the only way to cope with this problem is raising global logging level for metadataextraction mapper by manually setting:
```
<property>
    <name>mapred.map.child.log.level</name>
    <value>DEBUG</value>
</property>
```
in workflow definition.

We should change `DEBUG` logging level to `INFO` in all relevant places.

marekhorst:
Changing logging level. This should be piece of cake ;)


mkobos:
LGTM

PR #166, opened 2015-12-28
Closes #165: fix sentiment analysis precondition check

### LINKED ISSUE #165: fix sentiment analysis precondition check typo
Even though sentiment analysis is not integrated we do have integration test which fails after introducing invalid precondition check:
```
references.size() != sentiments.size()
```
we should check:
```
references.size() == sentiments.size()
```
instead.

marekhorst:
Another quick pull request :)

Again integration test proved to be quite useful.


mkobos:
LGTM

PR #169, opened 2016-01-07
Closes #94: version.properties contains git details instead of svn ones

### LINKED ISSUE #94: version.properties should contain git details instead of svn ones
After moving from svn to git `version.properties` file is empty:
```
svn.path=
svn.repository=
svn.revision=-1
```
In order to fix this we should provide git details to know which IIS revision was deployed.

One way could be to use `maven-git-commit-id-plugin`:
https://github.com/ktoso/maven-git-commit-id-plugin/

marekhorst:
After recent IIS deployment I realized we lack git details in `version.properties` file. This patch fixes this issue.

### DIFF
--- a/iis-workflows/pom.xml
+++ b/iis-workflows/pom.xml
@@ -289,18 +289,25 @@
 									</includePropertyKeysFromFiles>
 								</configuration>
 							</execution>
+						</executions>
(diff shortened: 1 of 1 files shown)


mkobos:
LGTM

Issue #170, opened 2016-01-12
document text collapser fails occasionally on large input plaintexts


marekhorst:
We can mitigate this problem in at least two supplementary ways:

* put size limit on XML/plaintext content read into IIS in the same way we limit PDF contents size
* override maximum XMX memory allocated to union PIG scripts. All mapreduce collapsers have this memory parameter adjusted we should do the same for PIG union script

marekhorst:
After further workflow inspection it turns out we could introduce another processing optimization: instead of collapsing two large `DocumentText` datastores (one with all texts extracted from all PDF and XML resources, the other with texts extracted from webcrawl) we could run two research-initiative algorithms on individual inputs and collapse outputs.

This should be much faster and less memory related error prone.

PR #173, opened 2016-01-13
#Closes 170: documenttext collapser fails with oomerr

marekhorst:
This one consists of 3 commits, each one improves stability.


mkobos:
Dividing this pull request into three functional commits makes reviewing much easier!

mkobos:
Commit messages are a nice documentation of the introduced change. So if eventually you want to squash the commits into a single one, its message in the second and later lines should contain this original documentation. On the other hand, one might consider leaving the original commits as they are (with potentially adding a prefix to the message of the intermediary commits like "Issue #170" so it's obvious which issue they are addressing).

marekhorst:
I've just added some comments and committed reformatted workflow.xml files.

mkobos:
Ok. There are two comments that require your answer: https://github.com/openaire/iis/pull/173#commitcomment-15613122, https://github.com/openaire/iis/pull/173#commitcomment-15688427

mkobos:
Ok. What about comment https://github.com/openaire/iis/pull/173#commitcomment-15613122?

mkobos:
LGTM

PR #177, opened 2016-01-14
Closes #176: supplement integration test suite

### LINKED ISSUE #176: supplement integration test suite with missing test cases
There are cases where integration test is defined only as `workflow.xml` file without `WorkfowTest` class triggering its execution. This was caused by the previous approach where integration tests were defined in `deploy.info` file making `WorkflowTest` class redundant.

Since moving our code to git and changing the way integration tests are triggered we need to reintroduce `WorkflowTest` classes with test methods triggering `workflow.xml` execution.

Here is the full list of orphan workflows:
```
eu/dnetlib/iis/workflows/ingest/pmc/metadata/sampledataproducer
eu/dnetlib/iis/workflows/ingest/pmc/plaintext/sampledataproducer
eu/dnetlib/iis/workflows/ingest/pmc/plaintext/brokenxml
eu/dnetlib/iis/workflows/ingest/pmc/plaintext/nulltext

eu/dnetlib/iis/workflows/top/integration/import
eu/dnetlib/iis/workflows/top/integration/primary/processing
```

marekhorst:
Some integration test `workflow.xml` definitions were not hooked up with `WorkflowTest` classes.

### DIFF
--- a/iis-workflows/iis-workflows-ingest-pmc/src/test/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/WorkflowTest.java
+++ b/iis-workflows/iis-workflows-ingest-pmc/src/test/java/eu/dnetlib/iis/workflows/ingest/pmc/metadata/WorkflowTest.java
@@ -1,0 +1,22 @@
+package eu.dnetlib.iis.workflows.ingest.pmc.metadata;
+
+import org.junit.Test;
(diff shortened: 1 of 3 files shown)


mkobos:
LGTM.

BTW, to save you some effort in the future - you could have skipped adding any comment to this pull request since its goal is already well described in issue #176.

PR #183, opened 2016-01-28
Closes #182: fix workflows-top integration test

### LINKED ISSUE #182: fix workflows-top integration test
`workflows-top` integration test fails with the following message:
```
E0743: Multiple "ok to" transitions to the same node, [decision-statistics], are not allowed
```
which seems to be caused by improper handling of `decision` nodes in `fork`s, reported as [OOZIE-1403](https://issues.apache.org/jira/browse/OOZIE-1403).

The way to overcome this issue is to set:
```
oozie.wf.validate.ForkJoin=false
```
in `job.properties` file.

Since our integration tests refactoring and blacklisting all `job.properties` files, including the ones associated with integration tests, we need to find other way to pass this parameter.

marekhorst:


mkobos:
LGTM

Issue #195, opened 2016-02-17
support JATS records nested inside OAI elements in both PMC metadata and plaintext ingesters


marekhorst:
This issue is related to integration of frontiers XML records described in details in openaire issue tracker:

[redmine:1660](https://issue.openaire.research-infrastructures.eu/issues/1660)

Each JATS record is nested inside oai `metadata` element.

PR #198, opened 2016-02-19
Marekhorst 195 support jats records nested inside oai elements

marekhorst:


mkobos:
I wasn't able to see what is exactly the functionality that is being implemented here and this made my work as a reviewer harder. I think that there should have been some high-level description (say 3-sentences long) placed somewhere - either in the description of the original issue #195 or in the description of the pull request or (ideally) in the commit comment. There's a link to a Redmine issue in the description of the issue #195 but the Redmine issue is long so I think it would be too demanding for me to read it.

marekhorst:
I have pushed another commit.

mkobos:
> This is the core change - supporting JATS records nested in OAI elements.
> 
> There are two possible scenarios:
> 
>     article is nested inside oai metadata element
>     article is root element
> 
> Second scenario was the only one supported up until now.

This comment is precious. I guess that it should be placed either somewhere in the code or in the commit comment.

mkobos:
Please separate changes to indentation into a separate commit next time.

mkobos:
Some minor comments. Apart from them, LGTM.

marekhorst:
> This comment is precious. I guess that it should be placed either somewhere in the code or in the commit comment.

I will put it in commit comment.

marekhorst:
> Please separate changes to indentation into a separate commit next time.

I am always trying not to forget about it. Maybe we should create dedicated task for fixing indentation once and for good, so we won't have to deal with this at each single issue?

mkobos:
> > Please separate changes to indentation into a separate commit next time.
> 
> I am always trying not to forget about it. Maybe we should create dedicated task for fixing indentation once and for good, so we won't have to deal with this at each single issue?

Seems to be a good idea.

PR #233, opened 2016-03-07
Marekhorst 67 core examples structure refactoring

marekhorst:


madryk:
I thought a little different about this task.
For easier understanding lets look at: `eu/dnetlib/iis/core/examples/java/cloner/oozie_app/workflow.xml`
For me `<action name="producer">` in this workflow is not part of example - it's there only to get data needed for testing purposes. Same goes for json files.
So this should stay in `src/test`.
Only `<action name="cloner">` is part of example.
But from description of task #67 implicates that you did it right. I don't know if I misunderstood the whole concept.
Did you discussed in detail about this task with Łukasz?

marekhorst:
You are right in a way if we want to make it in best possible way we should extract core module into separate workflow, move it to `src/main` and define it as a subworkflow in test workflow. The problem is the amount of time this would take: looking at the number of examples it looks like a tedious job for a week or so.

We didn't discuss this task in details with Lukasz but since we agreed it should take half a day I assume this was the job to be done.

madryk:
Ok. Then it seems to be ok

PR #240, opened 2016-03-10
Madryk 190 Write in spark output transformer for citation matching

madryk:


pom.xml
```diff
@@ -411,6 +411,10 @@
                         <groupId>org.apache.hadoop</groupId>
                         <artifactId>hadoop-core</artifactId>
                     </exclusion>
+                    <exclusion>
+                        <groupId>org.mortbay.jetty</groupId>
+                        <artifactId>servlet-api-2.5</artifactId>
```
marekhorst: Why to do we need that?

marekhorst:
LGTM.

PR #268, opened 2016-03-23
Closes #261: Separate metadataextraction and ingest_pmc models

### LINKED ISSUE #261: Separate metadataextraction and ingest_pmc models

lukdumi:


madryk:
Everything looks OK.
But I don't think that `rawText` should be computed from other affiliation metadata in `CermineAffiliation`. IMO `rawText` shouldn't be even a part of `CermineAffiliation`.
`rawText` should be some kind of unprocessed text so we would not loose any information that was maybe wrongly processed by cermine.
Maybe we could discuss about it and implement it differently in future as part of another task (current task you can merge as it is).

lukdumi:
OK, I've had create a separate issue for this #271.

PR #270, opened 2016-03-25
Marekhorst 236 update classification algorithm resources

marekhorst:


marekhorst:
Yesterday Ioannis confirmed integration tests results are expected so we are ready to merge updated documents classification algorithm with `master` branch.

lukdumi:
Marek, I don't understand why you have removed the tests instead of changing them.

PR #275, opened 2016-03-29
Closes #192: metadataextraction top level workflow

### LINKED ISSUE #192: Create metadataextraction top level workflow
To be triggered by DNet after harvesting new PDF contents from repository.

It should definitely wrap ```top/metadataextraction/cached_by_checksum``` workflow, possibly by extending ```top/common/import``` utilizing metadataextraction part only.

marekhorst:
I've split it into 2 commits: 

- cache building workflow based on `metadataextraction_cached` workflow but focusing on writing new data into cache rather than retrieving anything
- performance improvement: replacing original `skip_extracted` (which is still in use in cached metadataextraction) with better performing `skip_extracted_without_meta`. This PIG subworkflow is better suited for building cache scenario because it doesn't retrieve anything from cache.


marekhorst:
I just noticed typos in commit comments: issue number was set to 193 instead of 192. I will fix it while squashing commits during merge.

iis-wf/iis-wf-top/src/main/resources/eu/dnetlib/iis/wf/top/cache_builder/oozie_app/workflow.xml
```diff
@@ -0,0 +1,499 @@
...
+				</property>
+            	<property>
+					<name>input</name>
+					<value>${workingDir}/transformers_metadataextraction_skip_extracted/tobeprocessed_content</value>
+				</property>
+				<property>
+					<name>output_root</name>
+					<value>${workingDir}/metadata_extractor/output_root</value>
+				</property>
+				<!-- all the other properties are autmatically propagated-->
```
madryk: Problem with indentation

iis-wf/iis-wf-transformers/src/main/resources/eu/dnetlib/iis/wf/transformers/metadataextraction/skip_extracted_without_meta/oozie_app/lib/scripts/transformer.pig
```diff
@@ -0,0 +1,24 @@
...
+'schema', '$schema_document_content');
+
+documentContent = load '$input_document_content' using avro_load_document_content;
+documentMeta = load '$input_document_meta' using avro_load_document_meta;
+
+cachedDocumentId = foreach documentMeta generate id;
+cachedDocumentIdDistinct = distinct cachedDocumentId;
+
+joinedDocumentContent = join documentContent by id left, cachedDocumentIdDistinct by id;
+joinedFilteredDocumentContent = filter joinedDocumentContent by cachedDocumentIdDistinct::id is null;
```
madryk: It seems like left outer join is more than we need.
Inner join would be enough and you've could skip filter instruction that way.

iis-wf/iis-wf-transformers/src/test/resources/eu/dnetlib/iis/wf/transformers/metadataextraction/skip_extracted_without_meta/sampledataproducer/data/input_extracted_document_metadata.json
```diff
@@ -0,0 +1,3 @@
...
+{"publisher": "Harp3r T0rch", "affiliations": null, "authors": null, "language": "en", "title": "Small Gods", "externalIdentifiers": null, "journal": "Journal", "id": "XXXXX", "pages": null, "volume": "32", "references": [{"position": 1, "basicMetadata": {"publisher": null, "title": "The Lord of the Rings", "url": null, "series": null, "authors": null, "volume": null, "edition": null, "source": null, "year": "2012", "issue": null, "pages": {"start": "1", "end": "236"}, "location": null, "externalIds": null}, "text": "J.R.R. Tolkien, The Lord of the Rings, 2012"}], "year": null, "keywords": null, "issue": "4", "abstract": null, "publicationTypeName": null, "text": "extracted text"}
```
madryk: For this test only id field is relevant. It would be a lot nicer if this field was defined at start of json records, not somwhere in the middle

iis-wf/iis-wf-top/src/main/resources/eu/dnetlib/iis/wf/top/cache_builder/oozie_app/workflow.xml
marekhorst: Fixed in 3rd commit.

iis-wf/iis-wf-transformers/src/test/resources/eu/dnetlib/iis/wf/transformers/metadataextraction/skip_extracted_without_meta/sampledataproducer/data/input_extracted_document_metadata.json
marekhorst: Fixed in 3rd commit.

iis-wf/iis-wf-transformers/src/main/resources/eu/dnetlib/iis/wf/transformers/metadataextraction/skip_extracted_without_meta/oozie_app/lib/scripts/transformer.pig
marekhorst: In fact we do need an outer join here because while joining tables A and B we are interested in subset of records appearing only in A, not in B (translating into this particular case: we are looking for contents given as input which were NOT stored in cache so they could be processed by `metadataextraction` algorithm).

The filter condition is `cachedDocumentIdDistinct::id is null` not `cachedDocumentIdDistinct::id is not null` what could be replaced by inner join.

madryk: Ok. You're right

PR #281, opened 2016-04-04
Closes #276: Import organizations from IS database

### LINKED ISSUE #276: Import organizations from IS database

lukdumi:


iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/converter/OrganizationConverter.java
```diff
@@ -0,0 +1,97 @@
...
+        
+        
+        return orgBuilder.build();
+        
+    }
+    
+
+    //------------------------ PRIVATE --------------------------
+    
+    private boolean isDataCorrect(Result source, OafEntity oafEntity) throws UnsupportedEncodingException {
```
marekhorst: I guess we don't need to pass this parameter since it is not validated in method's body.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/converter/OrganizationConverter.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Converts {@link Oaf} object containing {@link eu.dnetlib.data.proto.OrganizationProtos.Organization}
+     * into {@link Organization}
+     */
+    @Override
+    public Organization buildObject(Result source, Oaf resolvedOafObject) throws UnsupportedEncodingException {
+        
+        
+        Preconditions.checkNotNull(resolvedOafObject);
+        Preconditions.checkNotNull(source);
```
marekhorst: We shouldn't care about source in this context. If `resolvedOafObject` is not null we are fine.

iis-wf/iis-wf-top/src/test/resources/eu/dnetlib/iis/wf/top/integration/import/oozie_app/workflow.xml
```diff
@@ -34,6 +34,15 @@
 					<name>output</name>
 					<value>${workingDir}/mr_import</value>
 				</property>
+               <property>
```
marekhorst: Yeah, I stuck on the very same issue this morning while testing importer integration test after moving it to `iis-wf-import` module.

This means integration test was never run on CDH5 infrastructure as it requires providing zookeeper properties explicitly. CDH4 cluster was configured in a way we didn't have to bother about those values, it was taken from environment which is much easier and convenient.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/converter/OrganizationConverter.java
lukdumi: Yes, you're right (I had used it for a while and then changed the method without removing it from the parameter list).

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/converter/OrganizationConverter.java
lukdumi: :+1:

lukdumi:
I've also changed the coansys version (proper changes in coansys has been merged into the master branch)

iis-wf/iis-wf-top/src/test/resources/eu/dnetlib/iis/wf/top/integration/import/oozie_app/workflow.xml
lukdumi: ok

PR #305, opened 2016-04-06
Closes #285: Fix avrostorage in pig scripts around cache workflows

### LINKED ISSUE #285: Built-in AvroStorage causes "Can't redefine: authors" pig error when manipulating data in cache
Since replacing piggybank version of `AvroStorage` with built-in one in #265 all cache manipulating PIG scripts throw:
```
ERROR 2116: Can't redefine: authors
```
error when writing avro datastore containing [ExtractedDocumentMetadata](https://github.com/openaire/iis/blob/5c090408fa5d89fd1a68223c1db8af715e09526f/iis-schemas/src/main/resources/eu/dnetlib/iis/metadataextraction/ExtractedDocumentMetadata.avdl) records.

It seems it is related to `authors` field name ambiguity and I assume the following fields are causing this problem:
- `ExtractedDocumentMetadata#authors`
- `ExtractedDocumentMetadata#references#basicMetadata#authors`

I didn't find any other candidates.

I am not sure why pig consider this as a problem, probably data modelling on pig layer is unaware of field record inheritance and this is when the error occurs. `piggybank` didn't have problem with this anyway.

If we are going to fix this simply by renaming one of the fields we will be forced to regenerate cache from scratch. But it seems this problem affects cdh5 branch only so this shouldn't be major problem since large cache was generated only on cdh4 cluster. We can sacrifice cdh5 cache.

madryk:


iis-wf/iis-wf-transformers/src/main/resources/eu/dnetlib/iis/wf/transformers/metadataextraction/checksum/postprocessing/meta/oozie_app/lib/scripts/transformer.pig
```diff
@@ -21,7 +21,7 @@ outputDocumentMeta = foreach joined generate
 		sourceDocumentMeta::year as year,
 		sourceDocumentMeta::publisher as publisher,
 		sourceDocumentMeta::references as references,
-		sourceDocumentMeta::authors as authors,
+		sourceDocumentMeta::authors as authorsMeta,
```
marekhorst: Good to see that - it means problem was related to pig names only and we don't need to rename field in avro schema.

Does postprocessing integration test confirm `authors` are properly written? It seems `outputDocumentMeta` field name is `authorsMeta` from now on, is AvroStorage capable of matching it properly with `authors` field defined in avro schema?

iis-wf/iis-wf-transformers/src/test/resources/eu/dnetlib/iis/wf/transformers/metadataextraction/checksum/preprocessing/sampledataproducer/data/output_document_content_url.json
```diff
@@ -1 +1 @@
-{"id": "12345", "url": "http://someurl1.com", "mimeType": "pdf", "contentChecksum": "12345", "contentSizeKB": 0}
\ No newline at end of file
+{"id": "12345", "url": "http://someurl2.com", "mimeType": "pdf", "contentChecksum": "12345", "contentSizeKB": 0}
```
marekhorst: This worries me, because it means outcome is somehow indeterminate. Even though both links should point to the same content (due to the same checksum).

marekhorst:
I guess there is one another place affected with the same issue:

https://github.com/openaire/iis/blob/e5af8bb49d0b932810c0c8f4eef780ce55e56266/iis-wf/iis-wf-transformers/src/main/resources/eu/dnetlib/iis/wf/transformers/metadataextraction/skip_extracted/oozie_app/lib/scripts/transformer.pig

Strangely I don't see integration tests for this one - we should add it. Just crated #306 issue for this.

iis-wf/iis-wf-transformers/src/main/resources/eu/dnetlib/iis/wf/transformers/metadataextraction/checksum/postprocessing/meta/oozie_app/lib/scripts/transformer.pig
madryk: In test for this transformer we have documentMetadata with some authors and it seems like output is correct.
AvroStorage most probably ignores field names and depends on order of fields in case when schema is defined directly in `AvroStorage('$schema_extracted_document_metadata');`

iis-wf/iis-wf-transformers/src/test/resources/eu/dnetlib/iis/wf/transformers/metadataextraction/checksum/preprocessing/sampledataproducer/data/output_document_content_url.json
madryk: Well, it's hard for me to determine if this is problematic in terms of already existing cache.
But there is not much we can do now about it. This change probably happened due to change of pig or some hadoop component (when we switched from cdh4 to cdh5)

marekhorst: No, it's not. It is only a problem with verifying test results.

madryk:
Ok. I will correct `metadataextraction/skip_extracted` pig script

iis-wf/iis-wf-transformers/src/test/resources/eu/dnetlib/iis/wf/transformers/metadataextraction/checksum/preprocessing/sampledataproducer/data/output_document_content_url.json
madryk: Well, in this test, output data relies on order of records read from datastore and how pig operations are constructed.

madryk: Actually I looked at pig documentation of `LIMIT` and encounter on not good info:
`There is no guarantee which n tuples will be returned, and the tuples that are returned can change from one run to the next`
so we have to fix this test

madryk: I've decided to order records by id before limit operation. This way test should be deterministic.

PR #310, opened 2016-04-07
Closes #290: Run oozie launchers in different queue in tests

### LINKED ISSUE #290: Resolve the problem with forking in oozie workflows

madryk:


iis-common/src/test/resources/integration-test-default.properties
```diff
@@ -29,3 +29,5 @@ oozie.execution.log.file.location=target/extract-and-run-on-remote-host.log
 
 # location of maven executable
 maven.executable=mvn
+
+oozieLauncherQueueName=default_oozie_launcher
```
marekhorst: Those are default properties so we don't need to provide it explicitly e.g. on jenkins worker node, right?

marekhorst:
I wasn't able to review it with full attention to every change - just scanned through all workflows and pom.xml file.

iis-common/src/test/resources/integration-test-default.properties
madryk: Yes. No change in jenkins will be needed.

PR #339, opened 2016-04-19
Closes #336: Correct ingest-pmc workflow tests

### LINKED ISSUE #336: Check possibility of importing only coansys document-similarity workflow definition
Currently we import whole oozie package from coansys.
We could decrease size of `iis-wf-documentssimilarity` oozie package (and projects that depends on it) by importing workflow definition and java classes separately. Then libraries that are used in both coansys and iis could be defined only once in oozie package. We could then exclude libraries that are provided on cluster (hadoop libs) from coansys document-similarity.

madryk:


marekhorst:
Just apply `ctrl`+`shift`+`f` and fix indentation in xml files.

PR #350, opened 2016-04-21
Closes #73: Prepare tests verifying exporter module serializing actions to json

### LINKED ISSUE #73: Prepare tests verifying exporter module serializing actions to json
Original issue related to exporter module was already reported in #37 and pull request handled at #57.

We are still missing tests verifying whether exporting process works as expected. Due to close deadlines we couldn't incorporate it into #37.

marekhorst:
You should find more details in commit comments.

Generally I've split this rather large task in 3 commits, each one is self-sufficient and can be analysed individually.


iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+import org.apache.hadoop.io.Writable;
+import org.apache.hadoop.util.ReflectionUtils;
+import org.apache.log4j.Logger;
+
+/**
+ * Iterator that extracts sequence file's consecutive {@link Text} values. 
+ * 
+ * @author mhorst
+ */
+public class SequenceFileTextValueReader implements CloseableIterator<Text> {
```
mpol: Sections :-(

mpol:
Code sections :-(

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+        private SequenceFile.Reader sequenceReader;
+
+        private final RemoteIterator<LocatedFileStatus> fileIt;
+
+        private final FileSystem fs;
+
+        /** Ignore file starting with underscore. Such files are also ignored by
+         * default by map-reduce jobs.
+         */
+        private final static String whitelistRegexp = "^(?!_).*";
```
mpol: Precompile with `Pattern.compile`?
Also "^[^_].*" could be more understandable at first sight (and only differ in not matching an empty string).

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+
+        private Text toBeReturned;
+
+        /**
+         * Default constructor.
+         * 
+         * @param fs hadoop filesystem
+         * @param seqPath sequence file path
```
mpol: Actual params differ.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+                        fileIt = fs.listFiles(path.getPath(), false);
+                        sequenceReader = getNextSequenceReader();
+                } else {
+                        fileIt = null;
+                        sequenceReader = new Reader(fs.getConf(),
+                                        SequenceFile.Reader.file(path.getPath()));
+                }
+        }
+
+        final Reader getNextSequenceReader() throws IOException {
```
mpol: private?

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        /**
+         * Checks whether file is valid candidate.
+         * @param fileStatus
+         * @return true when valid, false otherwise
+         */
+        private final boolean isValidFile(LocatedFileStatus fileStatus) {
+                if (fileStatus.isFile()) {
+                        if (whitelistRegexp!=null) {
```
mpol: if (WHITELIST_REGEXP != null)
Isn't checking a constant for null too defensive?

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+                                                SequenceFile.Reader.file(currentFileStatus.getPath()));
+                        }
+                }
+//              fallback
+                return null;
+        }
+
+        /**
+         * Checks whether file is valid candidate.
+         * @param fileStatus
```
mpol: description missing

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+                                return true;
+                        }
+                }
+//              fallback
+                return false;
+        }
+
+        /**
+         * Returns next data package.
+         * @return next data package
```
mpol: At least one of these lines seems unnecessary.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+         * @throws IOException
+         */
+        protected Text getNext() {
```
mpol: private?

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+                        }
+                }
+//              fallback
+                return false;
+        }
+
+        /**
+         * Returns next data package.
+         * @return next data package
+         * @throws IOException
```
mpol: Doesn't throw IOException.
(Aside: maybe CloseableIterator should be like/extend RemoteIterator?)

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+                        if (sequenceReader==null) {
+                                return null;
+                        }
+                        Writable key = (Writable) ReflectionUtils.newInstance(
+                                        sequenceReader.getKeyClass(), fs.getConf());
+                        Writable value = (Writable) ReflectionUtils.newInstance(
+                                        sequenceReader.getValueClass(), fs.getConf());
+                        if (sequenceReader.next(key, value)) {
+                        return (Text) value;
+                      } else {
```
mpol: Indentation!

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+                          sequenceReader.close();
+                          sequenceReader = getNextSequenceReader();
+                          if (sequenceReader!=null) {
```
mpol: `getNext` handles this case.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+                        toBeReturned = getNext();
+                } 
+                return toBeReturned!=null;
+        }
+
+        /* (non-Javadoc)
+         * @see java.util.Iterator#next()
+         */
+        @Override
+        public Text next() {
```
mpol: Should throw `NoSuchElementException`

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+         * @see eu.dnetlib.iis.exp.iterator.ClosableIterator#close()
+         */
+        @Override
+        public void close() {
+                if (sequenceReader!=null) {
+                        try {
+                                sequenceReader.close();
+                        } catch (IOException e) {
+                                log.error("error occurred when closing sequence reader", e);
+                        }
```
mpol: Why are we eating the exception here?

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
```diff
@@ -0,0 +1,168 @@
...
+
+        /*
+         * (non-Javadoc)
+         * 
+         * @see java.util.Iterator#remove()
+         */
+        @Override
+        public void remove() {
+                throw new UnsupportedOperationException();
+        }
```
mpol: No need to override `remove` if the implementation is the same as the default.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
mpol: No unit tests.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
```diff
@@ -0,0 +1,143 @@
...
+
+	private static final String GETTER = "get";
+
+	private Map<String, FieldDecoder> decoders = new HashMap<String, FieldDecoder>();
+
+	/**
+	 * Registers field decoder.
+	 * 
+	 * @param field
+	 * @param decoder
```
mpol: Remove or complete tags.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
```diff
@@ -0,0 +1,143 @@
...
+	 */
+	public void registerDecoder(String field, FieldDecoder decoder) {
+		decoders.put(field, decoder);
+	}
+
+	/**
+	 * Extracts value from given object based on fieldPath.
+	 * @param fieldPath field names acessible vie getter methods, nested fields are separated with dots
```
mpol: "accessible"

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
```diff
@@ -0,0 +1,143 @@
...
+	 * @param object object value should be extracted from
+	 * @return extracted value
+	 * @throws Exception
+	 */
+	public Object getValue(String fieldPath, Object object) throws Exception {
+		try {
+			if (object != null && fieldPath != null) {
```
mpol: Do we really want to not throw for `null` arguments?

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
```diff
@@ -0,0 +1,143 @@
...
+				Object currentObject = object;
+				String[] tokens = StringUtils.split(fieldPath, FIELD_PATH_SEPARATOR);
+				for (String token : tokens) {
+					if (token.charAt(0) == DECODER_PREFIX) {
+						String fieldName = token.substring(1);
+						FieldDecoder decoder = decoders.get(fieldName);
+						if (decoder == null) {
+							throw new Exception("no decoder registered for " + token);
```
mpol: Throw something more specific, or at least `RuntimeException`

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
```diff
@@ -0,0 +1,143 @@
...
+						}
+						Object fieldValue = getFieldValue(fieldName, currentObject);
+						if (decoder.canHandle(fieldValue)) {
+							currentObject = decoder.decode(fieldValue);
+						} else {
+							throw new Exception(fieldValue!=null?
```
mpol: `RuntimeException`

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
```diff
@@ -0,0 +1,143 @@
...
+ * keys. Nested field names are separated with dots.
+ * 
+ * Notice: field names should start with capital letters as they appear in getter method.
+ * 
+ * All fields requiring decoder should be marked with '$' prefix. 
+ * 
+ * @author mhorst
+ *
+ */
+public class FieldAccessor {
```
mpol: Couldn't the class be simpler (and not blow the 20 lines/method limit) if it used PropertyUtils from Apache Commons?

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
```diff
@@ -0,0 +1,143 @@
...
+					} else if (token.indexOf('[')>0) {
+						String fieldName = token.substring(0, token.indexOf('['));
+						currentObject = getFieldValue(fieldName, currentObject);
+						if (currentObject!=null) {
+							 if (currentObject.getClass().isArray()) {
+								 currentObject = Array.get(currentObject, getArrayPositionFromToken(token));	 
+							 } else if (List.class.isAssignableFrom(currentObject.getClass())) {
+								 currentObject = ((List<?>)currentObject).get(getArrayPositionFromToken(token));
+							 } else {
+								 throw new Exception("unable to extract array or list element "
```
mpol: `RuntimeException`

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
```diff
@@ -0,0 +1,143 @@
...
+											+ "based on field path: " + fieldPath + ", "
+											+ "object is neither an array nor list:" + 
+											currentObject.getClass().getName());
+							 }
+						} else {
+							throw new Exception("unable to extract array or list element "
```
mpol: `RuntimeException` (maybe NPE)

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	/**
+	 * Extracts value from given object based on fieldPath.
+	 * @param fieldPath field names acessible vie getter methods, nested fields are separated with dots
+	 * @param object object value should be extracted from
+	 * @return extracted value
+	 * @throws Exception
+	 */
+	public Object getValue(String fieldPath, Object object) throws Exception {
```
mpol: Doesn't handle "$decoded[indexed]" fields, which is probably OK, but could be expected if not documented.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
```diff
@@ -0,0 +1,143 @@
...
+						currentObject = getFieldValue(token, currentObject);
+					}
+				}
+				return currentObject;
+			} else {
+				return null;
+			}
+		} catch (Exception e) {
+//			we are extremely verbose to make log analysis process easier 
+			throw new Exception("unable to resolve field path '" + fieldPath
```
mpol: `RuntimeException`

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: Done.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: Done.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: This might be challenging due to `FileSystem`. I will get back to this.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: Why not? Done.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: Done.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: Yeah, it's dull. I have resurrected `SequenceFileTextValueReader` from old svn revision (it was dropped when we switched to avro), probably in first version of this class regexp was optional and after changing it to constant this condition was not removed.

Fixed.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
```diff
@@ -0,0 +1,143 @@
...
+		} catch (Exception e) {
+//			we are extremely verbose to make log analysis process easier 
+			throw new Exception("unable to resolve field path '" + fieldPath 
+					+ "' on object " + object, e);
+		}
+	}
+
+	/**
+	 * Extracts array/list position from given token
+	 * @param token
```
mpol: Remove or complete tags.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
```diff
@@ -0,0 +1,143 @@
...
+	private static Object getFieldValue(String fieldName, Object object) throws Exception {
+		Method method = getMethod(object.getClass(), GETTER + fieldName);
+		return method.invoke(object);
+	}
+
+	/**
+	 * Provides {@link Method} for given class and method name
+	 * @param clazz class from which method should be returned
+	 * @param name method name
+	 * @return declated method
```
mpol: "declared"?

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
```diff
@@ -0,0 +1,143 @@
...
+	 * @throws NoSuchFieldException
+	 * @throws NoSuchMethodException
+	 */
+	private static Method getMethod(Class<?> clazz, String name) throws NoSuchFieldException, NoSuchMethodException {
```
mpol: `clazz.getMethod()` would work for public getters. Do we need to find non-public getters?

mpol:
In addition to the code sections, the subsections in tests are missing as well.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: Done.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessorTest.java
```diff
@@ -0,0 +1,145 @@
+package eu.dnetlib.iis.wf.export.actionmanager.sequencefile;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.fail;
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.cxf.helpers.IOUtils;
```
mpol: Do we want such a dependency here? Apache Commons IOUtils could be more convenient.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessorTest.java
```diff
@@ -0,0 +1,145 @@
...
+		final String[] names;
+		public ArrayWrapper(String[] names) {
+			this.names = names;
+		}
+		public String[] getNames() {
+			return names;
+		}
+	}
+	
+	class ListWrapper {
```
mpol: `static class`

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessorTest.java
```diff
@@ -0,0 +1,145 @@
...
+	@Test(expected=Exception.class)
+	public void testAccessingListElement() throws Exception {
+		FieldAccessor accessor = new FieldAccessor();
+		ListWrapper object = new ListWrapper(Arrays.asList(new String[] {"name-1", "name-2"}));
+		assertEquals("name-1", accessor.getValue("Names[0]", object).toString());
+		assertEquals("name-2", accessor.getValue("Names[1]", object).toString());
+		accessor.getValue("Names[2]", object);
+	}
+	
+	class ArrayWrapper {
```
mpol: `static class`

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessorTest.java
```diff
@@ -0,0 +1,145 @@
...
+				.getResourceAsStream("/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/resultProject.json")),
+				"utf8"));
+		action.setTargetValue(decodedTargetValue);
+
+		assertEquals(rawSetId, accessor.getValue("RawSet", action).toString());
+		assertEquals(agentId, accessor.getValue("Agent.Id", action).toString());
+		assertEquals(agentName, accessor.getValue("Agent.Name", action).toString());
+		assertEquals(Agent.AGENT_TYPE.service.toString(), accessor.getValue("Agent.Type", action).toString());
+		assertEquals("relation", accessor.getValue("$TargetValue.Kind", action).toString());
+		assertEquals("resultProject", accessor.getValue("$TargetValue.Rel.RelType", action).toString());
```
mpol: All the toStrings should be unnecessary, I think.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessorTest.java
```diff
@@ -0,0 +1,145 @@
...
+		FieldAccessor accessor = new FieldAccessor();
+		accessor.registerDecoder("TargetValue", new OafFieldDecoder());
+
+		String rawSetId = "rawset-id";
+		String agentId = "agent-id";
+		String agentName = "agent-name";
+		AtomicAction action = new AtomicAction(rawSetId, new Agent(agentId, agentName, 
+				Agent.AGENT_TYPE.service));
+
+		accessor.getValue("$TargetValue.Unknown", action).toString();
```
mpol: Does it test what it should test? After changing "Unknown" to "Kind" the test still passes.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: > Doesn't throw IOException.

Fixed.

> (Aside: maybe CloseableIterator should be like/extend RemoteIterator?)

Yes, we could since `remove()` is never supported. But I would prefer to do this in separate task because this would require changes in several places where we currently expect `java.util.Iterator` instance (e.g. `eu.dnetlib.iis.wf.referenceextraction.dataset.DatasetDBBuilder` and `eu.dnetlib.iis.wf.referenceextraction.project.ProjectDBBuilder`). In this task we should also check whether it is ok to expect plain `java.util.Iterator` since it should be closed in the end...

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldDecoder.java
```diff
@@ -0,0 +1,26 @@
+package eu.dnetlib.iis.wf.export.actionmanager.sequencefile;
+
+/**
+ * Field decoder interface.
+ * @author mhorst
+ *
+ * @param <T>
```
mpol: unused parameter

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldDecoder.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public interface FieldDecoder {
+
+	/**
+	 * Decodes given source object into target representation.
+	 * @param source
```
mpol: Complete or remove

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldDecoder.java
```diff
@@ -0,0 +1,26 @@
...
+	 * @return decoded object
+	 * @throws Exception
+	 */
+	public Object decode(Object source) throws Exception;
+	
+	/**
+	 * Checks whether decoder can handle given input.
+	 * @param source
```
mpol: Complete or remove.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldDecoder.java
```diff
@@ -0,0 +1,26 @@
...
+	 * @return true when decoder can handle given input, false otherwise
+	 */
+	public boolean canHandle(Object source);
```
mpol: Is the expected semantics equivalent to
```
default boolean canHandle(Object source) {
    try {
        decode(source);
        return true;
    } catch (Exception e) {
        return false;
    }
}
```
or something else?

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/OafFieldDecoder.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ */
+public class OafFieldDecoder implements FieldDecoder {
+
+	/**
+	 * Decodes {@link Oaf} object from byte[].
+	 * @param source byte array
+	 * @return {@link Oaf} object
+	 * @throws InvalidProtocolBufferException
+	 * @throws UnsupportedEncodingException
```
mpol: Doesn't throw it.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/OafFieldDecoder.java
```diff
@@ -0,0 +1,37 @@
...
+	 */
+	@Override
+	public Oaf decode(Object source) throws Exception {
```
mpol: More precise `throws` declaration.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/OafFieldDecoder.java
```diff
@@ -0,0 +1,37 @@
...
+		builder.mergeFrom((byte[]) source);
+		return builder.build();
+	}
+
+	@Override
+	public boolean canHandle(Object source) {
+		return (source!=null && source.getClass().isArray() 
+				&& source.getClass().getComponentType().isPrimitive())
+				&& byte.class.isAssignableFrom(source.getClass().getComponentType());
+	}
```
mpol: `source != null && source.getClass().equals(byte[].class)`?

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/OrderedProperties.java
```diff
@@ -0,0 +1,104 @@
...
+	public boolean containsKey(Object key) {
+		return linkMap.containsKey(key);
+	}
+
+	public boolean containsValue(Object value) {
+		return linkMap.containsValue(value);
+	}
+
+	public Enumeration<Object> elements() {
+		throw new RuntimeException("Method elements is not supported in LinkedProperties class");
```
mpol: `return java.util.Collections.enumeration(linkMap.values())`

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: OK.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: Fixed.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/TestingConsumer.java
```diff
@@ -0,0 +1,120 @@
...
+				while (it.hasNext()) {
+					AtomicAction action = AtomicAction.fromJSON(it.next().toString());
+					actionsCount++;
+					if (actionsCount > recordsSpecs.length) {
+						throw new Exception("got more records than expected: " + "unable to verify record no " + actionsCount
+								+ ", no field specification provided! Record contents: " + action);
+					} else {
+						String currentSpecLocation = recordsSpecs[actionsCount - 1];
+						log.info("output specification location: " + currentSpecLocation);
+						Properties specProps = new OrderedProperties();
```
mpol: Why is the order so important?

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: This line is within `getNext()` method body...

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/TestingConsumer.java
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	@Override
+	public Map<String, PortType> getInputPorts() {
+		return inputPorts;
+	}
+
+	@Override
+	public Map<String, PortType> getOutputPorts() {
+		return new HashMap<String, PortType>();
```
mpol: `Collections.emptyMap()`?

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: True. Null won't be returned `NoSuchElementException` will be thrown instead.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/TestingConsumer.java
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	@Override
+	public void run(PortBindings portBindings, Configuration configuration, Map<String, String> parameters)
+			throws Exception {
+		String propertiesPathsCSV = parameters.get(PROPERTIES_CSV);
+		Path inputPath = portBindings.getInput().get(PORT_INPUT);
```
mpol: Switch the two lines above.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: Hmm, I don't recall why is this logged. 

Iterator does not seem to be the best place to log things anyway, added `throws` declaration instead.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
marekhorst: Removed.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/TestingConsumer.java
```diff
@@ -0,0 +1,120 @@
...
+		if (StringUtils.isEmpty(propertiesPathsCSV)) {
+			throw new Exception("no " + PROPERTIES_CSV + " property value provided, "
+					+ "output requirements were not specified!");
+		}
+		String[] recordsSpecs = StringUtils.split(propertiesPathsCSV, ',');
+		FileSystem fs = FileSystem.get(configuration);
+		if (fs.exists(inputPath)) {
```
mpol: Throw if not exists, fall through to the happy path otherwise. This puts precondition handling first and avoids some indentation.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/TestingConsumer.java
```diff
@@ -0,0 +1,120 @@
...
+			SequenceFileTextValueReader it = new SequenceFileTextValueReader(
+					new FileSystemPath(fs, inputPath));
+			try {
```
mpol: Use try-with-resources?

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: I knew someone has followed this path before me ;) I'll give it a try.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: > Remove or complete tags.

Detailed description added.

> Not obvious what exactly the field parameter should be. (A field name without the '$' prefix, I think.)

Exactly.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: Fixed.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: Hmm, for the sake of sanity I guess we should. Fixed.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: Done.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: Done.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: I can imagine decoder providing array or list as a result from which I would like to retrieve single element. So it should work, if it doesn't then there is a bug. I'll write unit test for this case.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: Handled.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: Handled.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: Handled.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: Description added.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: Fixed, although whole method will be removed when switching to `PropertyUtils`.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: True, `public` is enough. This will be probably replaced with `PropertyUtils`.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessorTest.java
marekhorst: Accidental eclipse pick, fixed.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessorTest.java
marekhorst: Removed all redundant occurrences.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessorTest.java
marekhorst: Yeap. This is because `$TargetValue` is enough for the test to pass as it is not present in action. What makes test fail is looking e.g. for `RawSet` or `Agent.Id` which is accessible.

I changed  `$TargetValue.Unknown` into `Agent.Unknown` to avoid confusions.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessorTest.java
marekhorst: ok

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessorTest.java
marekhorst: ok, in addition `public` also seem to be required by `PropertyUtils`.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldDecoder.java
marekhorst: Leftover from previous version, fixed.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldDecoder.java
marekhorst: Description added.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldDecoder.java
marekhorst: Description added.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldDecoder.java
marekhorst: Kind of. The initial intention was to allow using different decoder for the same `fieldName` and different type/context. 

This feature is not fully exploited by current `FieldAccessor` implementation yet.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/OafFieldDecoder.java
marekhorst: OK.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/OafFieldDecoder.java
marekhorst: Sticking to `InvalidProtocolBufferException`.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/OafFieldDecoder.java
marekhorst: Why not, seems to work, updated.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/OrderedProperties.java
marekhorst: This is external code but we could improve it :)

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/TestingConsumer.java
marekhorst: OK.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/TestingConsumer.java
marekhorst: OK.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/TestingConsumer.java
marekhorst: True, tabs saver :)

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/TestingConsumer.java
marekhorst: Right, at `cdh5` branch we are on `1.8` java compliance level so we can use it.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/TestingConsumer.java
marekhorst: From the technical POV it is not important. But when debugging failed test outcome it gives me additional information if I know that test failed at given line in properties file and I know all the expectation above this line were met.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: Finally I decided to introduce `FieldAccessorException` here and in all the places above. 

At first I didn't find justification for this (`TestingConsumer` treats all exceptions in the same way) but... it is always better to handle specific exceptions in upper layer, whatever it does. In future it can be utilized by other consumers which could do more with specific exception.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: `PropertyUtils` won't make this method significantly shorter so I have refactored this code into several specialized methods. To be committed...

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessor.java
marekhorst: I've just applied simple patch allowing subelement extraction from decoded object. Appropriate unit test was written for this case.

We do not support decoding array/list subelement though as it doesn't seem to be required.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
mpol: Yes. I mean this test is followed by a recursive call to `getNext()` which starts with an identical test.

marekhorst: `sequenceReader` is reassigned before second check what justifies its existence. It seems first `sequenceReader == null` check is currently required because `sequenceReader` is also instantiated by constructor and it might be `null`.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/TestingConsumer.java
mpol: > when debugging failed test outcome it gives me additional information if I know that test failed at given line in properties file and I know all the expectation above this line were met

True, although it suggests that maybe another format for the expectations (json array?) would be cleaner. But since it's test-only code and already written I guess it may stay as well.

iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java
mpol: > > (Aside: maybe CloseableIterator should be like/extend RemoteIterator?)
> 
> Yes, we could since remove() is never supported. But I would prefer to do this in separate task

Yes, of course.

marekhorst:
Let me add 100th comment in this thread :) 
I've just pushed post code-review commit with all mentioned fixes. All but one: unit test for `SequenceFileTextValueReader` - let me add dedicated ticket for this task (after merging this commit so I could point to the right resource). It doesn't look trivial and will take some time to develop.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessorTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+		assertEquals(rawSetId, accessor.getValue("rawSet", action));
+		assertEquals(agentId, accessor.getValue("agent.id", action));
+		assertEquals(agentName, accessor.getValue("agent.name", action));
+		assertEquals(Agent.AGENT_TYPE.service, accessor.getValue("agent.type", action));
+		assertEquals("relation", accessor.getValue("$targetValue.kind", action).toString());
+		assertEquals("resultProject", accessor.getValue("$targetValue.rel.relType", action).toString());
+	}
+
+	@Test(expected=Exception.class)
```
mpol: Now that we have `FieldAccessorException` we can be more precise in the tests.

mpol:
https://wiki.yadda.icm.edu.pl/yadda/Public:Class_code_sections

mpol:
For test subsections we don't seem to have any documented standard unless @lukdumi can point you to one. Your best bet is to look at some existing tests.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/FieldAccessorTest.java
marekhorst: Good idea, done.

marekhorst:
> For code sections: https://wiki.yadda.icm.edu.pl/yadda/Public:Class_code_sections

I've added sections comments. Although I didn't put sections in exceptions since there is nothing more than constructors. Same with `OrderedProperties` and `OafFieldDecoder` where there are no distinctive sections - logic only, so I guess there is no point.

marekhorst:
Please check my last commit.

PR #363, opened 2016-04-26
Closes #320: Ensure large files are used as metadata cache input

### LINKED ISSUE #320: Ensure large files are used as metadata cache input

mpol:
The actual metadata cache directory is usually (except on initialization)
created with Pig union, using a common union workflow.
As union is a map-only operation we cannot control the output file sizes
by changing the number of reducers, but we can use the pig.maxCombinedSplitSize
property to make Pig mappers combine the inputs.

Add a parameter for setting the property to the union workflow (defaulting to standard
Pig behaviour).
Use it in cache workflows to make union create large files, up to 320MB.

### DIFF
--- a/iis-wf/iis-wf-transformers/src/main/resources/eu/dnetlib/iis/wf/transformers/common/union/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-transformers/src/main/resources/eu/dnetlib/iis/wf/transformers/common/union/oozie_app/workflow.xml
@@ -13,6 +13,12 @@
 		<property>
 			<name>schema</name>
 		</property>
+		<property>
+			<name>combine_splits</name>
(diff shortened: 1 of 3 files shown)


iis-wf/iis-wf-transformers/src/main/resources/eu/dnetlib/iis/wf/transformers/common/union/oozie_app/workflow.xml
```diff
@@ -13,6 +13,12 @@
 		<property>
 			<name>schema</name>
 		</property>
+		<property>
+			<name>combine_splits</name>
+			<!-- zero will use the Pig default -->
+			<value>0</value>
```
marekhorst: 0 is neat. I was afraid we would need to provide current default value explicitly.

PR #385, opened 2016-05-11
Closes #364: Create SequenceFileTextValueReader unit test

### LINKED ISSUE #364: Create SequenceFileTextValueReader unit test
The following class:

https://github.com/openaire/iis/blob/cdh5/iis-common/src/main/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReader.java

is missing unit test.

First we need to decide whether we are able to create one since this reader is rooted into hadoop filesystem (`FileSystem` and `Reader` classes).

mpol:


iis-common/src/test/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReaderTest.java
```diff
@@ -0,0 +1,80 @@
...
+    @Test
+    public void readDirectory() throws Exception {
+        // given
+        SequenceFileTextValueReader reader = newReader(SEQUENCE_FILES);
+
+        // execute
+        ArrayList<String> items = readAll(reader);
+
+        // assert
+        assertThat(items, contains("a", "bb", "ccc", "a", "bb", "ccc"));
```
madryk: It would be nice if the second sequence file contained something else than the first one

iis-common/src/test/java/eu/dnetlib/iis/common/java/io/SequenceFileTextValueReaderTest.java
```diff
@@ -0,0 +1,80 @@
...
+    public void closeReader() throws Exception {
+        // given
+        SequenceFileTextValueReader reader = newReader(ONE_FILE);
+
+        // execute
+        reader.close();
+
+        // assert
+        thrown.expectCause(isA(IOException.class));
+        reader.hasNext();
```
madryk: You really test here that `hasNext` will fail after closing reader.
So i would name this test method `hasNext_closed_reader` or something similar.
Going further `reader.hasNext()` should be really in execute section (and `reader.close()` in given - because it is preparing the state of the service). But due to the way `ExpectedException` is constructed in junit, we cant really follow the rules of this 3 sections.
In conclusion I would exceptionally skip section comments in this particular test, as they are misleading.

mpol: It's more complicated than that. Maybe this test is not very useful?
`hasNext` doesn't have to throw after closing the iterator - in particular it will not throw if we are at the end of the iteration already, it will just return `false`, and that's OK.
What we are really interested in is that the `close` on reader does really close the underlying resources.
The exception from `hasNext` is only proxy for that. So in my mind it is really **given** a reader, **when** we close the reader, **assert** that to the best of our knowledge the underlying iterator was also closed.

madryk: Ok, lets leave it as is

PR #386, opened 2016-05-11
Closes #372: Change aff-match writer

### LINKED ISSUE #372: Adapt IisAffMatchResult writer to conform with exporter

madryk:


iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/write/BestMatchedAffiliationWithinDocumentPicker.java
```diff
@@ -0,0 +1,78 @@
...
+    
+    
+    //------------------------ PRIVATE --------------------------
+    
+    private Collection<List<MatchedAffiliation>> groupByOrganizationId(Iterable<MatchedAffiliation> matchedAffs) {
+        Map<CharSequence, List<MatchedAffiliation>> matchedGroupedByOrgId = Maps.newHashMap();
+        
+        for (MatchedAffiliation matched : matchedAffs) {
+            
+            if (!matchedGroupedByOrgId.containsKey(matched.getOrganizationId())) {
```
marekhorst: It is easier to read conditions without negations, so I would skip `!` and swap `if` and `else` bodies.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/write/BestMatchedAffiliationWithinDocumentPicker.java
```diff
@@ -0,0 +1,78 @@
...
+            
+        }
+        
+        return matchedGroupedByOrgId.values();
+    }
+    
+    private Tuple2<Integer, MatchedAffiliation> pickBetterMatch(Tuple2<Integer, MatchedAffiliation> first, Tuple2<Integer, MatchedAffiliation> second) {
+        
+        if (first._1 > second._1) {
+            return first;
```
marekhorst: Does it mean havining`OrgA` referenced twice in affliations with pretty poor `matchStrength` (which means it was accepted by rather insignificant voter) will be picked over `OrgB` referenced once with `matchStrength=1` (accepted by top voter)?

madryk: Yes. Not sure if this is the best approach.
I don't know right now how well this will behave. I would have to run aff-matching for bigger data.
Right now, when algorithm tends to assign low `matchStrength` for quite well matches it seemed like a good solution.
In future when algorithm will produce better `matchStrength` we could consider something different. Maybe comparing sums of `matchStrength` for matches with same `orgId`.
I don't really know.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/write/BestMatchedAffiliationWithinDocumentPicker.java
madryk: ok

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/write/BestMatchedAffiliationWithinDocumentPicker.java
marekhorst: It seems this solution is tailored for the current development phase in which we cannot put much trust on `matchStrength`. When we adjust this value to represent match strength accurately we should reimplement it in a way we trust more `matchStrength` not the number of matches. Can you create issue for this so we won't forget about it? It might be quite crucial in the end.

I've got this gut filling there is some paper referring to this kind of issue.

iis-wf/iis-wf-affmatching/src/main/resources/eu/dnetlib/iis/wf/affmatching/model/MatchedAffiliation.avdl
```diff
@@ -8,9 +8,6 @@ protocol IIS {
         // eu.dnetlib.iis.metadataextraction.schemas.ExtractedDocumentMetadata#id
         string documentId;
         
-        // position in the document (from 1)
-        int position;
-
```
marekhorst: Sorry for this late comment but I just noticed it: since we have removed `position` field it does not represent relation between affiliation and organization anymore, does it? It is more relation between document and organization. Maybe we should rename it into `DocumentOrganization` or `MatchedOrganization` which will be accurate in both cases ;)

madryk: Ok, I will rename it to `MatchedOrganization`

PR #437, opened 2016-06-01
Closes #419: Limit the load of the DNET services caused by IIS import

### LINKED ISSUE #419: Limit the load of the DNET services caused by IIS import

marekhorst:
by introducing support for dedicated ObjectStore importer queue


madryk:
What will happen when we run preprocessing or primary workflow?
Will `importerQueueName` be properly propagated?

marekhorst:
> What will happen when we run preprocessing or primary workflow?
> Will importerQueueName be properly propagated?

Yes, in the same way `queueName` is propagated: due to `<propagateConfiguration>` elements.

Issue #438, opened 2016-06-02
PB to Avro converters incorrectly expect unset optional fields to return null


mpol:
Observed at least in `PersonConverter`.

In fact the getters never return `null`

> If the field is not set, `getFoo()` will return a `Foo` instance with none of its fields set

-- https://developers.google.com/protocol-buffers/docs/reference/java-generated#singular-fields-proto2

so the logic in the converters is incorrect.

PR #442, opened 2016-06-02
Closes #391: Add missing unit tests for avro converters

### LINKED ISSUE #391: Add missing unit tests for avro converters
While preparing #373 I found some avro converter modules from `eu.dnetlib.iis.wf.importer.infospace.converter` package in `iis-wf-import` module lacking unit tests.

mpol:


iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/converter/PersonConverterTest.java
```diff
@@ -0,0 +1,82 @@
...
+        OafEntity oafEntity = emptyEntityBuilder(PERSON_ID).build();
+
+        // execute
+        Person person = converter.convert(oafEntity);
+
+        // assert
+        assertEquals(PERSON_ID, person.getId());
+        assertNull(person.getFirstname());
+        assertNull(person.getSecondnames());
+        assertNull(person.getFullname());
```
mpol: Here we assert the current behaviour, but it is probably not what is expected, because of #438.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/converter/OafRelToAvroConverterTestBase.java
```diff
@@ -0,0 +1,81 @@
...
+        assertEquals(SOURCE_ID, getSourceId.apply(rel));
+        assertEquals(TARGET_ID, getTargetId.apply(rel));
+    }
+
+    //------------------------ PRIVATE --------------------------
+
+    private OafRel createOafRelObject() {
+        return OafRel.newBuilder().setSource(SOURCE_ID).setTarget(TARGET_ID)
+                // the following values do not matter for the converters but the fields are required
+                .setRelType(irrelevantType).setSubRelType(irrelevantSubtype)
```
mpol: I wonder if setting these types for each test is a good idea at all. Maybe I should just put any sample types here? If the converters don't care, why should I?

marekhorst: We shouldn't care. Proper converter is picked based on `columnFamily` so there is no need to check `relType` or `subRelType`.

marekhorst:
I just looked at set of already implemented tests: `DocumentMetadataConverterTest` and `ProjectConverterTest` should be also supplemented with proper test cases.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/converter/PersonConverterTest.java
mpol: OK. So the test will have to change after fixing that bug. But for now it is what it is.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/converter/DocumentMetadataConverterTest.java
```diff
@@ -1,58 +1,275 @@
...
+    private static void addDescription(OafEntity.Builder builder, String value) {
+        builder.getResultBuilder().getMetadataBuilder().addDescriptionBuilder().setValue(value);
+    }
+
+    private static void setLanguage(OafEntity.Builder builder, String value) {
+        builder.getResultBuilder().getMetadataBuilder().getLanguageBuilder().setClassid(value);
+    }
+
+    private static ImmutableMap<String, List<QualifiedOafJsonRecord>> authorRelations() {
+        String jsonTemplate = "{\"kind\":\"relation\",\"rel\":{\"relType\":\"personResult\",\"subRelType\":\"authorship\",\"relClass\":\"hasAuthor\",\"source\":\"srcId\",\"target\":\"%s\",\"child\":false,\"personResult\":{\"authorship\":{\"ranking\":\"%s\",\"relMetadata\":{}}}}}";
```
marekhorst: We could read it from json resource. Not a strong requirement though.

PR #453, opened 2016-06-03
Closes #405: Drop hbase export support and simplify exporter module internal model

### LINKED ISSUE #405: Drop hbase export support and simplify exporter module internal model
We shouldn't support hbase export in CDH5 version. After dropping support for direct HBase export we could simplify `ActionBuilder` model.

marekhorst:


marekhorst:
Tons of old exporter code with fixed comments, indentations and sections.

marekhorst:
Just pushed two additional commits required after manual merge during rebasing.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/AbstractBuilderFactory.java
```diff
@@ -0,0 +1,67 @@
...
+import eu.dnetlib.iis.common.hbase.HBaseConstants;
+
+/**
+ * Abstract action builder factory.
+ * 
+ * @author mhorst
+ *
+ * @param <T> avro input type
+ */
+public abstract class AbstractBuilderFactory<T> implements ActionBuilderFactory<T> {
```
madryk: I think that `AbstractActionBuilderFactory` would be better name

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/AbstractBuilderFactory.java
```diff
@@ -0,0 +1,67 @@
...
+                .get(EXPORT_TRUST_LEVEL_THRESHOLD + EXPORT_ALGORITHM_PROPERTY_SEPARATOR + algorithmName.name());
+        if (algorithmTrustLevelThreshold != null
+                && !WorkflowRuntimeParameters.UNDEFINED_NONEMPTY_VALUE.equals(algorithmTrustLevelThreshold)) {
+            return Float.valueOf(algorithmTrustLevelThreshold);
+        }
+        String defaultTrustLevelThresholdStr = conf.get(EXPORT_TRUST_LEVEL_THRESHOLD);
+        if (defaultTrustLevelThresholdStr != null
+                && !WorkflowRuntimeParameters.UNDEFINED_NONEMPTY_VALUE.equals(defaultTrustLevelThresholdStr)) {
+            return Float.valueOf(defaultTrustLevelThresholdStr);
+        }
```
madryk: `WorkflowRuntimeParameters` have useful method `getParamValue` for retrieving params that are aware of `$UNDEFINED` constant. I would use it (after changing signature to use `Configuration` object instead of `JobContext`).
I would even consider to extend `WorkflowRuntimeParameters` to handle this whole situation (as separate method):
"Param was not found so use default param as fallback". I am pretty sure that this kind of getting params is used in couple places in iis.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
```diff
@@ -23,139 +24,142 @@
...
+    @Override
+    public ActionBuilderModule<Citations> instantiate(Configuration config, Agent agent, String actionSetId) {
+        return new CitationActionBuilderModule(provideTrustLevelThreshold(config), agent, actionSetId);
+    }
+
+    /**
+     * Creates {@link BlobCitationEntry} from {@link CitationEntry}.
+     * Translates confirence level into trust level applying confidenceToTrustLevelFactor.
+     */
+    public static BlobCitationEntry build(CitationEntry entry, float confidenceToTrustLevelFactor) {
```
madryk: method is used only in inner class so i would move it there

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/SequenceFileExporterMapper.java
```diff
@@ -35,135 +32,86 @@
...
+            return propertyValue;
+        } else {
+            // providing default actionset identifier
+            propertyValue = cfg.get(EXPORT_ACTION_SETID);
+            if (StringUtils.isNotBlank(propertyValue) && !WorkflowRuntimeParameters.UNDEFINED_NONEMPTY_VALUE.equals(propertyValue)) {
+                return propertyValue;
+            } else {
+                throw new MappingNotDefinedException(
+                        "no action set identifier defined " + "for algorithm: " + algorithmName.name());
+            }
```
madryk: `WorkflowRuntimeParameters.getParamValue`?

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/module/DocumentToSoftwareUrlActionBuilderModuleFactoryTest.java
```diff
@@ -48,43 +48,48 @@
...
     }
 
     // ----------------------- TESTS --------------------------
 
     @Test(expected = NullPointerException.class)
-    public void test_build_null_object() throws Exception {
+    public void test_constructor_null_agent() throws Exception {
         // execute
-        module.build(null, agent, actionSetId);
+        factory.new DocumentToSoftwareUrlActionBuilderModule(trustLevelThreshold, null, actionSetId);
```
madryk: This is unusual syntax. Just use `new DocumentToSoftwareUrlActionBuilderModule()`
This applies for every test for ActionBuilderModule.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
marekhorst: I guess it was left in factory because it is used in unit tests. This is the case when we change class internals due to external requirements (tests). Eventually we could move it to `CitationsActionBuilderModuleUtils` to make it more straightforward.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/module/DocumentToSoftwareUrlActionBuilderModuleFactoryTest.java
marekhorst: I agree, this is unusual but we are dealing with non-static inner class (relies on `AbstractActionBuilderFactory#algorithmName` which is non-static) so this construct is required.

We could drop this e.g. after making `ActionBuilderModule` `algorithmName` aware but on the other hand this would make each `ActionBuilderModule` more complex.

madryk: Ok, we can leave it unchanged.

PR #456, opened 2016-06-06
Closes #394: Write missing comments for aff-matching module

### LINKED ISSUE #394: Add comments and do other small refactorings in aff-matching
For example there is no comment above the method:
OrganizationSectionsSplitter.splitToSections or above the class SectionedNameLevenshteinMatchVoter

Other things:
- AbstractSectionedMatchVoter.voteMatch should be final
- containsMatchingSection - the name is too generic, it should say what it does, e.g. containsOrgSection(List<String> affOrgNameSections, String orgNameSection). This way the implementor of a new class will know what he should do without looking at the comment.

madryk:


iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/match/voter/CountryCodeLooseMatchVoter.java
```diff
@@ -20,6 +20,10 @@
     
     //------------------------ LOGIC --------------------------
     
+    /**
+     * Returns true if countryCode of affiliation and organization are equal
+     * or one of them is empty.
```
lukdumi: ... if countryCode**s**...
... or **at least** one of them is empty.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/match/voter/CountryCodeStrictMatchVoter.java
```diff
@@ -20,6 +20,9 @@
     
     //------------------------ LOGIC --------------------------
     
+    /**
+     * Returns true if countryCode of affiliation and organization are non empty and equal.
```
lukdumi: I think it would be better to use "link" more often:
... if {@link Affiliation#getCountryCode} and {@link AffMatchOrganization#getCountryCode} are non-empty and equal.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/match/voter/SectionedNameLevenshteinMatchVoter.java
```diff
@@ -20,31 +28,52 @@
     
     //------------------------ CONSTRUCTORS --------------------------
     
+    /**
+     * Default constructor
+     * 
+     * @param minSimilarity - minimum similarity for two sections to be found similar.
+     *      Value must be between (0,1] (minimum similarity equal to one means
+     *      that two sections must be equal).
+     */
```
lukdumi: I would also add:
@see #containsMatchingSection(...

PR #468, opened 2016-06-07
Closes #445: Remove obsolete citation matching related input parameters

### LINKED ISSUE #445: Remove obsolete citation matching related input parameters
Namely: `cit_genAuthorIdxJavaOpts` in primary `workflow.xml`. 

This parameter seems to be related to previous implementation of citation matching algorithm.

marekhorst:


marekhorst:
It turns out no more obsolete parameters were defined except the memory related one.

PR #471, opened 2016-06-08
Closes #458: Aff matching voter strength based on the test data processing instead of voter position

### LINKED ISSUE #458: Aff matching voter strength based on the test data processing instead of voter position

lukdumi:


iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/match/AffOrgMatcherFactory.java
```diff
@@ -97,12 +98,17 @@ public static AffOrgMatcher createDocOrgRelationMatcher(JavaSparkContext sparkCo
      * Creates {@link AffOrgMatchVoter}s for {@link #createDocOrgRelationMatcher(JavaSparkContext, String, String, String, Float)} 
      */
     public static ImmutableList<AffOrgMatchVoter> createDocOrgRelationMatcherVoters() {
+
+        FittingOrgWordsMatchVoter fitOrgWordsMatchVoter = new FittingOrgWordsMatchVoter(ImmutableList.of(','), 3, 0.8f, 0.9f);
```
marekhorst: Have you considered using spring or some other framework for parameterizing (and also wiring) beans?

Do you think there are any parameter candidates we should provide dynamically at runtime e.g. while conducting series of experiments on large input data.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/match/AffOrgMatcherFactory.java
```diff
@@ -153,11 +159,11 @@ public static AffOrgMatcher createMainSectionHashBucketMatcher() {
...
     public static ImmutableList<AffOrgMatchVoter> createMainSectionHashBucketMatcherVoters() {
         return ImmutableList.of(
-                createNameCountryStrictMatchVoter(),
-                createNameStrictCountryLooseMatchVoter(),
-                createSectionedNameStrictCountryLooseMatchVoter(),
-                createSectionedNameLevenshteinCountryLooseMatchVoter(),
-                createSectionedShortNameStrictCountryLooseMatchVoter());
+                createNameCountryStrictMatchVoter(1f),
+                createNameStrictCountryLooseMatchVoter(1f),
+                createSectionedNameStrictCountryLooseMatchVoter(0.914f),
```
marekhorst: Are those values provided based on accuracy on test data sample?

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/match/voter/AffOrgMatchVotersFactory.java
```diff
@@ -15,43 +15,62 @@ private AffOrgMatchVotersFactory() { }
...
     
-    public static AffOrgMatchVoter createNameCountryStrictMatchVoter() {
+    public static AffOrgMatchVoter createNameCountryStrictMatchVoter(float matchStrength) {
         
-        return new CompositeMatchVoter(ImmutableList.of(
+        CompositeMatchVoter voter = new CompositeMatchVoter(ImmutableList.of(
                 new CountryCodeStrictMatchVoter(), 
                 new NameStrictWithCharFilteringMatchVoter(ImmutableList.of(',', ';'))));
         
+        voter.setMatchStrength(matchStrength);
```
marekhorst: Is there any scenario we wouldn't like to set match strenght to `CompositeMatchVoter`? If no, maybe we could enforce providing it as constructor parameter?

Related to this line and all similar lines above.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/match/voter/FittingOrgWordsMatchVoter.java
```diff
@@ -121,11 +121,21 @@ private String filterName(String name) {
...
         }
         
         return filteredName;
     }
+
+    //------------------------ toString --------------------------
+    
+    @Override
+    public String toString() {
+        return "FittingOrgWordsMatchVoter [charsToFilter=" + charsToFilter
```
marekhorst: You may consider using either guava's `MoreObjects.toStringHelper()` or `ToStringBuilder` from `commons-lang`.

iis-wf/iis-wf-affmatching/src/test/java/eu/dnetlib/iis/wf/affmatching/match/voter/AffOrgMatchVoterStrengthEstimator.java
```diff
@@ -149,12 +140,25 @@ public static void classCleanup() throws IOException {
     @Test
     public void estimateVoterMatchStrengths_for_DocOrgRelationMatcher() throws IOException {
         
+        // given
+        
+        createInputDataFromJsonFiles(
+                ImmutableList.of("src/test/resources/experimentalData/input/all_organizations.json"),
```
marekhorst: `src/test/resources/experimentalData/input/` base path appears in 42 places in all affmatch tests, we could consider extracting this root path to some shared test variable.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/match/AffOrgMatcherFactory.java
lukdumi: As to the spring, yes, I thought about it but I'm not sure if it is a good idea, and certainly, not here along with this task.
And as far as parameters are concerned, no, I don't think we can do something in real time. Do you have any ideas?

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/match/AffOrgMatcherFactory.java
lukdumi: Yes.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/match/voter/AffOrgMatchVotersFactory.java
lukdumi: Well, probably not, but there are other Voters that don't have it always set. And passing it sometimes in constructors and sometimes via setters would make a mess.

iis-wf/iis-wf-affmatching/src/test/java/eu/dnetlib/iis/wf/affmatching/match/voter/AffOrgMatchVoterStrengthEstimator.java
lukdumi: :+1:

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/match/voter/FittingOrgWordsMatchVoter.java
lukdumi: I'll have to think about it, I've never used these methods before.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/match/AffOrgMatcherFactory.java
marekhorst: > As to the spring, yes, I thought about it but I'm not sure if it is a good idea, and certainly, not here along with this task.

OK, just wanted to indicate using spring seems to be tempting. Not in this task, of course.

> And as far as parameters are concerned, no, I don't think we can do something in real time. Do you have any ideas?

All the hardcoded flat/int values attracted my attention (e.g. the ones parameterizing `FittingOrgWordsMatchVoter`) - if they are going to be changed regularly while adjusting algorithm it might be reasonable to provide them at runtime between executions. If not - lets leave it as it is.

lukdumi: Well, we could of course generate it in mvn or something, and maybe one day we will do it. But for now it seems to me this is not very important thing. And this will only change when the test data changes and it won't very often.
But you're right, the best solution would be for it to work automatically.

PR #498, opened 2016-06-15
Closes #487: Allow multiple aff matches for single affiliation

### LINKED ISSUE #487: Change the algorithm so it returns multiple aff matches
Every affiliation should go through all matchers.

madryk:


iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/AffMatchingService.java
```diff
@@ -124,23 +128,26 @@ private void checkArguments(JavaSparkContext sc, String inputAffPath, String inp
...
         
         
         for (AffOrgMatcher affOrgMatcher : affOrgMatchers) {
             
-            JavaRDD<AffMatchResult> matchedAffOrgs = affOrgMatcher.match(idAffiliations.values(), normalizedOrganizations);
+            JavaRDD<AffMatchResult> matchedAffOrgs = affOrgMatcher.match(normalizedAffiliations, normalizedOrganizations);
             
-            allMatchedAffOrgs = allMatchedAffOrgs.union(matchedAffOrgs);
+            JavaPairRDD<String, AffMatchResult> matchedAffOrgsWithKey = matchedAffOrgs
+                    .keyBy(x -> x.getAffiliation().getId() + AFF_ORG_ID_SEPARATOR + x.getOrganization().getId());
```
mpol: I would probably use Tuple2<String, String> as key.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/match/AffMatchResultChooser.java
```diff
@@ -13,7 +13,7 @@
  * @author Łukasz Dumiszewski
 */
 
-class AffMatchResultChooser implements Serializable {
+public class AffMatchResultChooser implements Serializable {
```
mpol: This is now always called with equal organization ids. Should be simplified, or not?

PR #504, opened 2016-06-17
Closes #502: Adjust executor parameters for affmatch and citmatch in …

### LINKED ISSUE #502: Adjust executor parameters for affmatch and citmatch in primary integration test
Both `affiliationmatching` and `citationmatching` workflows have set of spark executor related parameters explicitly set: `sparkExecutorMemory`, `sparkExecutorCores`, `sparkExecutorsNumber` which are set by default to the values appropriate to OCEAN cluster configuration.

In `affiliationmatching` and `citationmatching` integration tests we override those default values with the ones tailored for our test cluster:
```
sparkExecutorMemory=1G
sparkExecutorCores=1
sparkExecutorsNumber=2
```
we have to do the same for `primary/processing` workflow integration test where both algorithms are triggered. We would need to extract those parameters up to the `primary/processing` workflow and override values at this point in integration test. Otherwise integration test hangs due to insufficient test cluster resources.

marekhorst:
…primary integration test

### DIFF
--- a/iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
@@ -322,6 +322,19 @@
 					<name>output_matched_doc_organizations</name>
 					<value>${workingDir}/exported/matched_affiliation</value>
 				</property>
+                <!-- shared executors configuration for affmatch and citmatch spark algorithms-->
+                <property>
+                    <name>sparkExecutorMemory</name>
+                    <value>1G</value>
+                </property>
+                <property>
+                    <name>sparkExecutorCores</name>
+                    <value>1</value>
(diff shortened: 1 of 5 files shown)


iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
```diff
@@ -322,6 +322,19 @@
...
+                <property>
+                    <name>sparkExecutorMemory</name>
+                    <value>1G</value>
+                </property>
+                <property>
+                    <name>sparkExecutorCores</name>
+                    <value>1</value>
+                </property>
+                <property>
+                    <name>sparkExecutorsNumber</name>
```
lukdumi: The same here, could we change the sparkExecutorsNumber to sparkNumExecutors to match the prop name as the other vars?

marekhorst:
Just renamed all `sparkExecutorsNumber` occurrences into `sparkNumExecutors` in my last commit.

PR #524, opened 2016-06-22
Closes #523: Integration test fails for direct citation matching

### LINKED ISSUE #523: Integration test fails for direct citation matching
After #506 the integration test fails on the test cluster because Spark requests too much memory from the cluster and the job hangs without executors.

mpol:


iis-wf/iis-wf-citationmatching-direct/src/test/resources/eu/dnetlib/iis/wf/citationmatching/direct/test/oozie_app/workflow.xml
```diff
@@ -77,6 +77,10 @@
                     <name>output</name>
                     <value>${workingDir}/output/citation</value>
                 </property>
+                <property>
+                    <name>sparkExecutorMemory</name>
+                    <value>1G</value>
+                </property>
```
madryk: It would be nice to limit sparkNumExecutors also. Let's say to 2 executors?

PR #525, opened 2016-06-22
Closes #497: Extend and fill organizations with alternative names

### LINKED ISSUE #497: Extend the aff matching module with alternative names field and fill it during reading
Use the institution map created in #429

madryk:


iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/AffMatchingJob.java
```diff
@@ -112,12 +118,31 @@ private static AffMatchingService createAffMatchingService(JavaSparkContext spar
...
+        affMatchingService.setAffMatchOrganizationAltNameFiller(altNameFiller);
         
         return affMatchingService;
     }
     
+    private static AffMatchOrganizationAltNameFiller createAffMatchOrganizationAltNameFiller() throws IOException {
+        
+        AffMatchOrganizationAltNameFiller altNameFiller = new AffMatchOrganizationAltNameFiller();
+        
+        List<String> altNamesCountryCodes = of(
```
marekhorst: I am wondering how dynamic this list could become. If it is rather static - let's leave it as it is. But if it is going to be changed from time to time then maybe we should define it somewhere alongside files with mappings (as a let's call it registry file). This way someone adding new language file will spot registry file right away and add new entry in registry. Otherwise one would need to know there is java definition to be altered and recompiled.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/AffMatchingJob.java
```diff
@@ -112,12 +118,31 @@ private static AffMatchingService createAffMatchingService(JavaSparkContext spar
...
-        affMatchingService.setAffOrgMatchers(of(docOrgRelationMatcher, mainSectionHashBucketMatcher, firstWordsHashBucketMatcher));
+        AffMatchOrganizationAltNameFiller altNameFiller = createAffMatchOrganizationAltNameFiller();
+        affMatchingService.setAffMatchOrganizationAltNameFiller(altNameFiller);
         
         return affMatchingService;
     }
     
+    private static AffMatchOrganizationAltNameFiller createAffMatchOrganizationAltNameFiller() throws IOException {
+        
+        AffMatchOrganizationAltNameFiller altNameFiller = new AffMatchOrganizationAltNameFiller();
```
marekhorst: It's only after I read this name for the 10th time I finally realized it is `Filler` not `Filter`. I was going to ask what do we filter here ;)

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/model/AffMatchOrganization.java
```diff
@@ -78,6 +82,34 @@ public String getWebsiteUrl() {
...
+    public void addAlternativeName(String alternativeName) {
+        this.alternativeNames.add(alternativeName);
+    }
+    
+    /**
+     * Returns the list with all organization names.
+     * That is main {@link #getName()} with all {@link #getAlternativeNames()} 
+     */
+    public List<String> getNames() {
+        List<String> names = Lists.newArrayList(name);
```
marekhorst: I guess it will be better to allocate `ArrayList` with fixed capacity (`alternativeNames.size() +1`) and then add all the names. This will prevent from potential reallocation.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/orgalternativenames/AffMatchOrganizationAltNameFiller.java
```diff
@@ -0,0 +1,85 @@
...
+    
+    
+    //------------------------ PRIVATE --------------------------
+    
+    private List<String> fetchAlternativeNames(List<String> originalNames) {
+        
+        for (List<String> alternativeNames : alternativeNamesDictionary) {
+            
+            if (containsSameOrganizationName(originalNames, alternativeNames)) {
+                return filterAlreadyPresentAlternativeNames(alternativeNames, originalNames);
```
marekhorst: Do we want to interrupt after finding first match? 

If we are dealing with original English name then we will stop after finding first match from the first language package, am I right?

I am wondering about covering the scenario when affiliation contains English organization name while organization name is defined in some other language. Was there any case of finding non-english name among set of organizations?

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/orgalternativenames/CsvOrganizationAltNamesDictionaryFactory.java
```diff
@@ -0,0 +1,60 @@
...
+    
+    //------------------------ LOGIC --------------------------
+    
+    /**
+     * Creates dictionary of alternative organization names using csv files located 
+     * under provided classpaths.
+     */
+    public List<List<String>> createAlternativeNamesDictionary(List<String> classpathCsvResources) throws IOException {
+        
+        List<List<String>> dictionary = Lists.newArrayList();
```
marekhorst: Have you considered using `List` of `Set` of `String`s? This will eliminate potential duplicates right away.

iis-wf/iis-wf-affmatching/src/test/java/eu/dnetlib/iis/wf/affmatching/AffMatchingAffOrgQualityTest.java
```diff
@@ -253,6 +255,29 @@ private AffMatchingService createAffMatchingService() {
...
+        
         return affMatchingService;
     }
+   
+   
+   private AffMatchOrganizationAltNameFiller createAffMatchOrganizationAltNameFiller() throws IOException {
+       
+       AffMatchOrganizationAltNameFiller altNameFiller = new AffMatchOrganizationAltNameFiller();
+       
+       List<String> altNamesCountryCodes = of(
```
marekhorst: Again, if we had it in resource file we wouldn't need to define set of languages explicitly. I guess we could even be more generic and do not assume those are languages and simply put file names in registry. 

Going even further we could drop `universities_` prefix from file name since we have it all in `universities` directory.

iis-wf/iis-wf-affmatching/src/test/java/eu/dnetlib/iis/wf/affmatching/match/voter/AffOrgMatchVoterStrengthEstimator.java
```diff
@@ -405,9 +408,30 @@ private AffMatchingService createAffMatchingService() {
...
+        affMatchingService.setAffMatchOrganizationAltNameFiller(altNameFiller);
+        
         return affMatchingService;
     }
     
+    private AffMatchOrganizationAltNameFiller createAffMatchOrganizationAltNameFiller() throws IOException {
+        
+        AffMatchOrganizationAltNameFiller altNameFiller = new AffMatchOrganizationAltNameFiller();
+        
+        List<String> altNamesCountryCodes = of(
```
marekhorst: My other comments related to `altNamesCountryCodes` are applicable here as well.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/AffMatchingJob.java
madryk: Indeed, they are very similar. I can't think of any better name though.
If you have some proposition. Please tell.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/orgalternativenames/CsvOrganizationAltNamesDictionaryFactory.java
madryk: This is a good idea

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/orgalternativenames/AffMatchOrganizationAltNameFiller.java
madryk: Not sure if I understand it right.

There is no need to always go through all alternativeNames. We assume that single organization alternative names are defined in one line and only in single file. The same organization can't be added to dictionary twice with different alternative names. So first match is good enough. Otherwise some merging mechanism should be implemented when building dictionary.

Yes, we always stop after finding first match (in any of the alternative names). We're looking in all of alternative names so it doesn't matter if its English or native.

If affiliation is in English and organization is in native language. Then to organization alternative names English version will be added. Next we will have to update voters, so they will use alternativeName field. And match will be found.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/AffMatchingJob.java
madryk: IMO it is overcomplication.
If someone is adding new file with alternative names pack in `src/main/resources/eu/dnetlib/iis/wf/affmatching/universities` then it is implied that he knows about definition in java code because he had to somehow find this files. Assuming standard approach for exploration of the code (that is top-down approach) developer will go through `workflow.xml` ->`AffMatchingJob` -> `createAffMatchOrganizationAltNameFiller()` and he would find location of the all alternative names files. Next he would think: 'All the files are under `src/main/resources/eu/dnetlib/iis/wf/affmatching/universities` classpath. Ok, I will add new file there`.
As for definition existing in several places - I will fix it.

iis-wf/iis-wf-affmatching/src/test/java/eu/dnetlib/iis/wf/affmatching/AffMatchingAffOrgQualityTest.java
madryk: OK for defining resources as simple list - maybe it will be more readable if there will be no loop for creating classpath strings.
I would leave `universities_` prefix in files. It is not long and quite 'readable'. If I would see bunch of files named with just two letters my first thought would be: Those files are named with some two random letters. When filename is prefixed then I would think that they are not some random and my brain would caught quicker that those two letters stands for country code.

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/orgalternativenames/AffMatchOrganizationAltNameFiller.java
marekhorst: > BTW. universities resource files are not really language packages it's rather country packages (pl stands for Poland in universities_pl.csv not Polish)

OK, now it's all clear!

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/AffMatchingJob.java
marekhorst: Appender, attacher? Supplier maybe? We could stick with `Filler` - since I finally read what was written I know what it does ;)

iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/orgalternativenames/OrganizationAltNameConst.java
```diff
@@ -0,0 +1,44 @@
...
+            "/eu/dnetlib/iis/wf/affmatching/universities/universities_lt.csv",
+            "/eu/dnetlib/iis/wf/affmatching/universities/universities_lu.csv",
+            "/eu/dnetlib/iis/wf/affmatching/universities/universities_lv.csv",
+            "/eu/dnetlib/iis/wf/affmatching/universities/universities_nl.csv",
+            "/eu/dnetlib/iis/wf/affmatching/universities/universities_pl.csv",
+            "/eu/dnetlib/iis/wf/affmatching/universities/universities_pt.csv",
+            "/eu/dnetlib/iis/wf/affmatching/universities/universities_ro.csv",
+            "/eu/dnetlib/iis/wf/affmatching/universities/universities_se.csv",
+            "/eu/dnetlib/iis/wf/affmatching/universities/universities_si.csv",
+            "/eu/dnetlib/iis/wf/affmatching/universities/universities_sk.csv"
```
marekhorst: Nice. We could also extract common parts of all those paths into some `rootPath` variable. But I don't want to be too picky so it's all up to you.

PR #550, opened 2016-07-11
Closes #549: Build report for info space import count entities

### LINKED ISSUE #549: Collect entities counters from import from info space
That is number of:
dedupmap, docmeta, docproj, organization, person, project, projectorg

madryk:


iis-common/src/main/java/eu/dnetlib/iis/common/report/ReportGenerator.java
```diff
@@ -0,0 +1,107 @@
...
+        
+        
+        FileSystem fs = FileSystem.get(conf);
+        
+        Path reportPath = portBindings.getOutput().get(REPORT_PORT_OUT_NAME);
+        
+        DataStore.create(avroReportParams, new FileSystemPath(fs, reportPath));
+        
+    }
+
```
marekhorst: PRIVATE section.

iis-wf/iis-wf-import/src/main/resources/eu/dnetlib/iis/wf/importer/infospace/oozie_app/workflow.xml
```diff
@@ -75,6 +75,11 @@
             <value>projectorg</value>
             <description>project-organization output name</description>
         </property>
+        
+        <property>
+            <name>output_report</name>
+            <description>main directory for storing partial reports</description>
```
marekhorst: I'm not sure we should call it partial reports. It depends on the context, from the whole processing POV those results are partial but from this importer POV those are just reports.

iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/main/oozie_app/workflow.xml
```diff
@@ -490,6 +490,10 @@
                     <name>output_faults</name>
                     <value>${workingDir}/primary_import/faults</value>
                 </property>
+                <property>
+                    <name>output_report</name>
+                    <value>${workingDir}/report</value>
```
marekhorst: For now this would work but considering we are going to produce counters for various importer modules we need more fine-grained reports directory structure along with joiners. We are going to produce report datastores in several places within importer subworkflow (metadata importer which is already covered but also content importer, dataset etc) and we wound need to join them before providing them to uber workflow (caller).

madryk: My intention was to store all these reports in the same parent folder.
Currently report for IS import will be located under `${workingDir}/report/import_info_space`. And for example report for datasets import can be located under `${workingDir}/report/import_dataset`.
This way joining all the partial reports can be located somewhere after export and we can provide just one parameter with this parent input location.
Enabling possibility to set partial report locations to "anywhere we want" causes a lot of juggling with path parameters.
In my opinion this is unnecessary.
We always use param `output_report` in workflows that reports something.
In primary workflow we only pass that `output_report` for `import`, `processing`, `export` subworkflows which is parent folder for all reports and let the `<propagate-configuration />` to do the rest. We don't have to modify anyhow middle-part workflows. And using only one param to control all the partial reports.

madryk: To me your proposition for joining of partial reports directly in parent workflow is also complicated more then it needs to be. This way we would end up with extra joining node in every middle workflow.

iis-wf/iis-wf-import/src/main/resources/eu/dnetlib/iis/wf/importer/infospace/oozie_app/workflow.xml
madryk: Ok, I agree with your statement.

madryk:
About clearing the situation that all the future report nodes will write to seperate folder.
Maybe it will be clearer if there would be two parameters in `/iis-wf-import/src/main/resources/eu/dnetlib/iis/wf/importer/infospace/oozie_app/workflow.xml` (and in future all the reporting workflows):

```
        <property>
            <name>output_report_base_path</name>
            <description>base directory for storing report</description>
        </property>
        <property>
            <name>output_report_relative_path</name>
            <value>import_info_space</value>
            <description>directory for storing report (relative to output_report_base path)</description>
        </property>

```
This way in theory we would have control over where finally report will be located. In practice though we will never be overriding `output_report_relative_path` value.
What do you think about that.

marekhorst:
Let's proceed as agreed on out Today's meeting.

PR #561, opened 2016-07-14
Closes #547: Make use of timeout param after upgrading CERMINE dependency to 1.9

### LINKED ISSUE #547: Make use of timeout param after upgrading CERMINE dependency to 1.9
We should upgrade CERMINE dependency to just released `1.9` version and set `timeout` parameter to prevent IIS from exceeding mapreduce heart beat timeout. This issue was described in:

https://github.com/CeON/CERMINE/issues/7

and fixed in:

https://github.com/CeON/CERMINE/pull/20

Currently `mapreduce.task.timeout` was set in `metadataextraction` workflow to ridiculous `3600000` value just to prevent whole IIS workflow from interruption when CERMINE stucks on some large document. 

Now we can rely on default `mapreduce.task.timeout` value (`600000` which is 10 minutes) and set CERMINE `timeout` value to something slightly less than that.

This should improve overall performance as well because we won't be blocked on CERMINE for more than 10 minutes. We could even do further analysis and see what is the CERMINE processing time distribution for all the contents and pick more appropriate value.

It will be extremely important to pick subset of PDF files causing CERMINE lags and run tests to make sure `timeout` parameter is properly adjusted so we do not exceed `mapreduce.task.timeout`.

marekhorst:


iis-wf/iis-wf-metadataextraction/src/main/java/eu/dnetlib/iis/wf/metadataextraction/AbstractMetadataExtractorMapper.java
```diff
@@ -138,153 +134,119 @@ protected void setup(Context context) throws IOException, InterruptedException {
...
-     * @param contentStream
-     * @throws IOException
-     * @throws InterruptedException
+     * @param documentId document identifier
+     * @param contentStream stream to be processed
      */
-    protected void processStream(CharSequence documentId, InputStream contentStream) throws IOException, InterruptedException {
+    protected void processStream(String documentId, InputStream contentStream) throws IOException, InterruptedException {
+        currentProgress++;
+        if (currentProgress > 0 && currentProgress % progresLogInterval == 0) {
```
madryk: `currentProgress` is always `>0`. So first condition can be removed

PR #565, opened 2016-07-18
Closes #544: Merge partial reports into single json file

### LINKED ISSUE #544: Write service that collects reports from submodules and combines them into final report

madryk:


iis-common/src/main/java/eu/dnetlib/iis/common/report/ReportParamJsonAppender.java
```diff
@@ -0,0 +1,95 @@
...
+        insertValue(jsonReport, Lists.newArrayList(jsonFieldHierarchy), convertParamValueToJson(reportParam));
+        
+    }
+    
+    
+    //------------------------ PRIVATE --------------------------
+    
+    private JsonElement convertParamValueToJson(ReportParam reportParam) {
+        
+        long value = Long.valueOf(reportParam.getValue().toString());
```
marekhorst: Now I am wondering, maybe we should keep this type aligned with `ReportParam#value` type? Either pass it as `string` or change `ReportParam#value` to `long`?

Otherwise someone generating partial report cound bump into troubles when setting `ReportParam#value` to something different than `string` representation of `float` value. Because the requirement of long type is hardcoded in this class only.

iis-common/src/main/java/eu/dnetlib/iis/common/report/ReportParamJsonAppender.java
```diff
@@ -0,0 +1,95 @@
...
+        
+        return new JsonPrimitive(value);
+    }
+    
+    
+    private void insertValue(JsonObject element, List<String> fieldHierarchy, JsonElement value) {
```
marekhorst: Even though this is private method it will be great to have at least brief description of what it does.

iis-common/src/test/java/eu/dnetlib/iis/common/java/jsonworkflownodes/TestingJsonConsumer.java
```diff
@@ -0,0 +1,102 @@
...
+ * @author madryk
+ */
+public class TestingJsonConsumer implements Process {
+
+    private final Map<String, JsonPortSpecification> inputPortsSpecification = Maps.newHashMap();
+    
+    
+    //------------------------ CONSTRUCTORS --------------------------
+    
+    public TestingJsonConsumer(String[] inputSpecifications) {
```
marekhorst: Can you describe in a comment what `inputSpecifications` property is?

iis-common/src/test/java/eu/dnetlib/iis/common/report/ReportMergerTest.java
```diff
@@ -0,0 +1,116 @@
...
+        // execute
+        
+        reportMerger.run(portBindings, conf, Collections.emptyMap());
+        
+        
+        // assert
+        
+        JsonObject actualReportJson = readJson(outputReportPath);
+        
+        String expectedReportJsonString = "{"
```
marekhorst: It is hard to maintain JSON or XML code defined in java. It will be easier having it defined in external file and reading it from classpath.

iis-common/src/test/java/eu/dnetlib/iis/common/report/ReportParamJsonAppenderTest.java
```diff
@@ -0,0 +1,125 @@
...
+    private JsonParser jsonParser = new JsonParser();
+    
+    
+    private JsonObject jsonReport;
+    
+    
+    @Before
+    public void setup() {
+        
+        String json = "{"
```
marekhorst: My previous JSON related comment applies here as well.

iis-common/src/test/java/eu/dnetlib/iis/common/report/ReportParamJsonAppenderTest.java
```diff
@@ -0,0 +1,125 @@
...
+        ReportParam reportParam = new ReportParam("property1.propertyB.propertyII", "21");
+        
+        // execute
+        
+        jsonAppender.appendReportParam(jsonReport, reportParam);
+        
+        // assert
+        
+        String expectedJson = "{"
+                + "  property1: {"
```
marekhorst: My previous JSON related comment applies here as well.

iis-common/src/test/java/eu/dnetlib/iis/common/report/ReportParamJsonAppenderTest.java
```diff
@@ -0,0 +1,125 @@
...
+        
+        ReportParam reportParam = new ReportParam("property1.propertyB", "21");
+        
+        // execute
+        
+        jsonAppender.appendReportParam(jsonReport, reportParam);
+        
+        // assert
+        
+        String expectedJson = "{"
```
marekhorst: My previous JSON related comment applies here as well.

iis-common/src/test/java/eu/dnetlib/iis/common/report/ReportParamJsonAppenderTest.java
```diff
@@ -0,0 +1,125 @@
...
+        
+        ReportParam reportParam = new ReportParam("property1.propertyA", "21");
+        
+        // execute
+        
+        jsonAppender.appendReportParam(jsonReport, reportParam);
+        
+        // assert
+        
+        String expectedJson = "{"
```
marekhorst: My previous JSON related comment applies here as well.

iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/main/oozie_app/workflow.xml
```diff
@@ -819,10 +819,30 @@
                 </property>
             </configuration>
         </sub-workflow>
-        <ok to="end" />
+        <ok to="build-report" />
         <error to="fail" />
     </action>
 
+    <action name="build-report">
```
marekhorst: Since we have `preprocessing` alongside `primary` workflow we should generate reports for both. Could you define it there as well? If you think this task deserves creating additional ticket, please do create one.

madryk: Ok. I will add it.

iis-common/src/main/java/eu/dnetlib/iis/common/report/ReportParamJsonAppender.java
madryk: I agree that this may be a little misleading.
But in future, it can be helpful.
If someone would like to use further the generated report. He would know that in this specific field he will always get a numeric value and would not have to parse this value.
I used `string` in `ReportParam.value` to be able to handle not only counters. When we will have different types of report params. For example times of workflows.
We could add a type enum field to ReportParam. And decode ReportParam.value based on this type.

PR #577, opened 2016-07-22
Closes #555: Report ingest pmc, metadataextraction and cache related …

### LINKED ISSUE #555: Collect counters from jobs related to cache, pmc and pdf faults

madryk:
…counters


iis-common/src/main/java/eu/dnetlib/iis/common/counter/PigCounters.java
```diff
@@ -0,0 +1,145 @@
...
+     */
+    public JobCounters getJobCounters(String jobId) {
+        return jobCountersMap.get(jobId);
+    }
+    
+    /**
+     * Returns job id with provided alias
+     * or <code>null</code> if job does not exists.
+     */
+    public String getJobIdByAlias(String jobAlias) {
```
lukdumi: This method is a bit strange. It returns the jobId of the <b>first</b> (no information given in the comment) encountered JobCounters which contains the given alias.

iis-common/src/main/java/eu/dnetlib/iis/common/report/PigCounterValueResolver.java
```diff
@@ -0,0 +1,65 @@
...
+            
+            String counterValue = pigCounters.getJobCounters(jobId).getCounter(counterName);
+            
+            if (StringUtils.isBlank(counterValue)) {
+                throw new IllegalArgumentException("Couldn't find a counter with name: " + counterName + " inside job counters, id: " + jobId);
+            }
+            return counterValue;
+        }
+        
+        return value;
```
lukdumi: Does it make any sense to return the passed value? Shoudn't an exception be thrown in such a case?

iis-common/src/main/java/eu/dnetlib/iis/common/report/PigCounterValueResolver.java
```diff
@@ -0,0 +1,65 @@
...
+    
+    /**
+     * Returns resolved value using provided {@link PigCounters}.<br/>
+     * Passed value can be a simple string or string with placeholder.
+     * Simple string will be returned unchanged.<br/>
+     * If value contains placeholder it will be resolved to corresponding
+     * value from {@link PigCounters}. Placeholder must be in form:
+     * <code>{JOB_ALIAS.JOB_COUNTER}</code>.
+     */
+    public String resolveValue(String value, PigCounters pigCounters) {
```
lukdumi: Wouldn't it be better for the method to be passed jobAlias and counterName? Why should one pass some strange placeholder string when one could pass just the two necessary values?

iis-common/src/main/java/eu/dnetlib/iis/common/report/PigCounterValueResolver.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import eu.dnetlib.iis.common.counter.PigCounters;
+
+/**
+ * Resolver of values with pig counters placeholder.
+ * 
+ * @author madryk
+ */
+public class PigCounterValueResolver {
```
lukdumi: This class looks weird. I guess it is focused on too many things. One thing is to parse some string into an alias and counter name and the other is to get from pigCounters a counter that is assigned the given alias and name.

iis-common/src/test/java/eu/dnetlib/iis/common/counter/PigCountersTest.java
```diff
@@ -0,0 +1,66 @@
...
+        
+        jobCounters1.addCounter("COUNTER_1", "34");
+        
+        jobCounters2 = new JobCounters("JOB_ID_2");
+        jobCounters2.addCounter("COUNTER_1", "13");
+        jobCounters2.addCounter("COUNTER_2", "vv");
+        
+        pigCounters = new PigCounters(Lists.newArrayList(jobCounters1, jobCounters2));
+    }
+
```
lukdumi: I would also add the cases that generate exceptions or return nulls.

iis-common/src/main/java/eu/dnetlib/iis/common/report/PigCountersReportGenerator.java
```diff
@@ -0,0 +1,116 @@
...
+    private Map<String, String> collectReportParameters(Map<String, String> parameters) {
+        
+        return parameters.entrySet().stream()
+                .filter(property -> property.getKey().startsWith(REPORT_PROPERTY_PREFIX))
+                .map(x -> Pair.of(x.getKey().substring(REPORT_PROPERTY_PREFIX.length()), x.getValue()))
+                .collect(Collectors.toMap(e -> e.getLeft(), e -> e.getRight()));
+        
+    }
+    
+    private List<ReportParam> convertToAvroReportParams(Map<String, String> reportParameters, PigCounters pigCounters) {
```
lukdumi: Wouldn't it be better to have it in a separate service? It would be simpler and easier to test.

iis-common/src/main/java/eu/dnetlib/iis/common/report/PigCountersReportGenerator.java
```diff
@@ -0,0 +1,116 @@
...
+    
+    @Override
+    public void run(PortBindings portBindings, Configuration conf, Map<String, String> parameters) throws Exception {
+        
+        String pigCountersJson = parameters.get(PIG_COUNTERS_PROPERTY);
+        
+        PigCounters pigCounters = pigCountersParser.parse(pigCountersJson);
+        
+        
+        Map<String, String> reportParams = collectReportParameters(parameters);
```
lukdumi: I would name this map something like 'reportCounterNames' and give some explanation that it is the map of pig counter names to human readable report names. It was hard for me to understand what it is.

lukdumi:
Krzyś, in general it's ok, but I'd like to talk a bit about it.

iis-common/src/main/java/eu/dnetlib/iis/common/counter/PigCounters.java
madryk: Aliases are unique in the whole pig job. So the first match is the only match.

iis-common/src/main/java/eu/dnetlib/iis/common/report/ReportPigCountersResolver.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public class ReportPigCountersResolver {
+
+    //------------------------ LOGIC --------------------------
+    
+    /**
+     * Resolve {@link ReportParam}s from {@link PigCounters} using {@link ReportPigCounterMapping}s.
+     * Only counters that are present in {@link ReportPigCounterMapping}s will be resolved.
+     */
+    public List<ReportParam> resolveReportCounters(List<ReportPigCounterMapping> reportPigCountersMapping, PigCounters pigCounters) {
```
lukdumi: Could we change the parameter order? I think it would be more natural to have pigCounters first. The method then could be read like this: "resolve report counters from pigCounters by using the given mapping".

PR #595, opened 2016-07-27
Closes #566: Report document texts count

### LINKED ISSUE #566: Counter of document texts extracted from ExtractedDocumentMetadata records
document texts extracted from ExtractedDocumentMetadata records (direct input to all referenceextraction algorithms)

madryk:


marekhorst:
Whole code looks OK but I am wondering whether we could prevent from resources duplication (e.g. PIG scripts, majority of wf definition etc).

One way to go could be making original union workflow counter aware but as optional feature. New functionality could be activated by conditional node when e.g. `union_count_report_key` was explicitly provided. If not - we could disable report generation action.

PR #610, opened 2016-08-03
Closes #604: Document classification reports - get rid of accumulators

### LINKED ISSUE #604: Document classification reports - get rid of accumulators

lukdumi:


iis-wf/iis-wf-documentsclassification/src/main/java/eu/dnetlib/iis/wf/documentsclassification/DocClassificationReportGenerator.java
```diff
@@ -0,0 +1,88 @@
...
+                                new ReportParam(ACM_CLASSES, ""+acmClassCount));
+    }
+    
+    private static long size(Collection<?> collection) { // static - spark will not have to serialize the DocClassificationReportGenerator object
+        
+        if (collection == null) {
+            return 0;
+        } 
+        
+        return collection.size();
```
marekhorst: We could also go with simple oneliner: `return collection!=null?collection.size():0;`

lukdumi: Well, this is a matter of taste. I don't think it's better, it's different.

PR #643, opened 2016-08-24
Closes #642: Some other corrections of execution report entry names

### LINKED ISSUE #642: Some other corrections of execution report entry names

lukdumi:


lukdumi:
I haven't changed 'milliseconds' to 'millisecond' because I'm not sure it it was a good solution. It looked strange and it would look even stranger if we decided to change it to 'second' ('second' could be understood as an adjective).

PR #658, opened 2016-09-09
Closes #656: Copy report to external localization

### LINKED ISSUE #656: Add support for copying report to external destination

madryk:


iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/main/oozie_app/workflow.xml
```diff
@@ -853,12 +858,16 @@
...
+                    <value>${workingDir}/report/working_dir</value>
+                </property>
+                <property>
                     <name>input_partial_reports</name>
                     <value>${workingDir}/report</value>
                 </property>
                 <property>
-                    <name>output_report</name>
-                    <value>${workingDir}/report/report.json</value>
+                    <name>reports_external_path</name>
```
marekhorst: This seems to be redundant when `<propagate-configuration />` is set.

iis-wf/iis-wf-preprocessing/src/main/resources/eu/dnetlib/iis/wf/preprocessing/main/oozie_app/workflow.xml
```diff
@@ -706,12 +711,16 @@
...
+                    <value>${workingDir}/report/working_dir</value>
+                </property>
+                <property>
                     <name>input_partial_reports</name>
                     <value>${workingDir}/report</value>
                 </property>
                 <property>
-                    <name>output_report</name>
-                    <value>${workingDir}/report/report.json</value>
+                    <name>reports_external_path</name>
```
marekhorst: This seems to be redundant when `<propagate-configuration />` is set.

iis-wf/iis-wf-preprocessing/src/main/resources/eu/dnetlib/iis/wf/preprocessing/main/oozie_app/workflow.xml
```diff
@@ -213,6 +213,11 @@
 			<value>/user/${user.name}/iis/working_dirs/${execution_environment}</value>
 			<description>working directory</description>
 		</property>
+        <property>
+            <name>reports_external_path</name>
+            <value>$UNDEFINED$</value>
```
marekhorst: Default value is quite crucial. It will be good not to require `reports_external_path` to be set if someone is not interested in gathering reports in external location.

Thus we should either provide here default location (e.g. `${workingDir}/report/report.json`) where report will be copied if external location is not desired.

Other approach could rely on `$UNDEFINED$` which is already set. But we would need to handle it properly in subworkflow handling copying report - e.g. by skipping copying when `reports_external_path=$UNDEFINED$`. Currently this is not the case, `$UNDEFINED$` directory will be created.

iis-wf/iis-wf-preprocessing/src/main/resources/eu/dnetlib/iis/wf/preprocessing/report/oozie_app/workflow.xml
```diff
@@ -37,6 +41,30 @@
...
+    
+    <action name="copy-report">
+        <distcp xmlns="uri:oozie:distcp-action:0.1">
+            <job-tracker>${jobTracker}</job-tracker>
+            <name-node>${nameNode}</name-node>
+            <prepare>
+                <mkdir path="${nameNode}${reports_external_path}" />
+            </prepare>
+            <arg>${nameNode}${working_dir}/report.json</arg>
+            <arg>${nameNode}${reports_external_path}/${wf:actionData('report_name_generator')['report_filename']}</arg>
```
marekhorst: I guess we could simplify filename generation by defining this argument as:
```
<arg>${nameNode}${reports_external_path}/preprocessing_report_${timestamp()}.json</arg>
```
You can find more details on `timestamp()` usage here:
http://stackoverflow.com/questions/21750151/getting-current-time-in-oozie-workflow

Date part is not that nice but it makes whole definition much simpler: we could get rid of `ReportFileNameGenerator` implementation and skip `report_name_generator` action execution.

What do you think?

marekhorst: One more comment: I just realised ":" is unsupported in paths:
https://issues.apache.org/jira/browse/HADOOP-3257

but we could look into other functions, e.g. on `replaceAll()` from workflow functional spec:
https://oozie.apache.org/docs/3.3.0/WorkflowFunctionalSpec.html
and make file name acceptable by `distcp` command.

iis-wf/iis-wf-preprocessing/src/main/resources/eu/dnetlib/iis/wf/preprocessing/main/oozie_app/workflow.xml
madryk: Ok. I will omit copying report when `reports_external_path=$UNDEFINED$`

iis-wf/iis-wf-preprocessing/src/main/resources/eu/dnetlib/iis/wf/preprocessing/report/oozie_app/workflow.xml
madryk: I considered using oozie `timestamp()` function.
Documentation says that it will return string like: `1997-07-16T19:20:30.45Z`. Which seemed too me quite long for a file name (seconds and fraction of seconds part).
I don't know. If you think using `timestamp()` will be better then I will correct it.

marekhorst: > I considered using oozie timestamp() function.
> Documentation says that it will return string like: 1997-07-16T19:20:30.45Z. Which seemed too me quite long for a file name (seconds and fraction of seconds part).
> I don't know. If you think using timestamp() will be better then I will correct it.

`timestamp()` alone is too long but consider using mentioned `replaceAll()` function.

In plain java:
```
System.out.println("1997-07-16T19:20:30.45Z".replaceAll("(..)T(..):(..)\\:(\\d)+\\.(\\d)+Z$", "$1_$2-$3"));
```
gives:
```
1997-07-16_19-20
```
which is exactly what we need. 

Assuming `replaceAll()` workflow function works the same  - we are there, without involving `ReportFileNameGenerator`.

madryk: As it turns out `${timestamp()}` returns only strings in form `2016-09-13T11:58Z`, so regex became simpler.

marekhorst:
One, very last thing: since `ReportFileNameGenerator` become obsolete we can remove both `ReportFileNameGenerator` and `ReportFileNameGeneratorTest` resources.

Please close this PR then and merge changes.

madryk:
Thanks. I completely forgot about that.

PR #752, opened 2017-02-20
RCUK support

sosguns2002:


sosguns2002:
Hello Marek, I just fixed the RCUK support

PR #769, opened 2017-03-15
NIH Optimization added

sosguns2002:
This optimization is general. Without this, every grant that is not from NIH will check for these rules that its not needed


sosguns2002:
Marek can you check if everything works well with this change? I forgot to do this optimization in the past

marekhorst:
I just merged this PR. It worked fine, results were not affected.

I assume I should expect performance improvement but probably due to `cache_builder` job ongoing in parallel no such improvement was noticed.

PR #789, opened 2017-05-11
add software mining enrichment from web

johnfouf:


marekhorst:
This pull request is outdated, new `software.sql` script version was introduced meanwhile.

PR #935, opened 2019-02-08
print error json record in stderr

johnfouf:


marekhorst:
Merged error propagation proposed in #936 PR.

PR #955, opened 2019-05-02
delete datacite records without publisher or creator names

johnfouf:


marekhorst:
Integrated with IIS master branch with f23fb653d618f492e2dd38215a4853204bdcfec6 commit.

PR #965, opened 2019-05-21
Add DFG version 2, update to solve some false positives

johnfouf:
…f another string


marekhorst:
Already merged with master branch with 6555bbd2f1eee93b4f9337f10803e1bcbdcc9437 commit.

PR #977, opened 2019-07-19
DARIAH-EU

johnfouf:


marekhorst:
#999 pull request will cover changes introduced in this PR.

PR #1001, opened 2019-11-12
add RSF

johnfouf:


marekhorst:
#1014 is the PR to be integrated.

PR #1005, opened 2019-11-21
Przemyslawjacewicz 973 integrate patents reference extraction with authors

przemyslawjacewicz:
This PR is about adding document authors to patents mining. It incorporates patent mining scripts as merged in #994. 

Additional fixes after patent mining job execution
1. explicit HDFS output locations removal in spark jobs
1. minor workflow xmls cleanup

Before merging this PR with base branch all commits should be squashed.


iis-wf/iis-wf-referenceextraction/src/main/java/eu/dnetlib/iis/wf/referenceextraction/patent/input/PatentReferenceExtractionInputDocumentTextAuthorMerger.java
```diff
@@ -0,0 +1,82 @@
...
+            JavaRDD<DocumentText> documentTexts = sparkAvroLoader.loadJavaRDD(sc, params.inputDocumentTextPath, DocumentText.class);
+            JavaRDD<ExtractedDocumentMetadataMergedWithOriginal> documentMetadatas =
+                    sparkAvroLoader.loadJavaRDD(sc, params.inputExtractedDocumentMetadataMergedWithOriginalPath, ExtractedDocumentMetadataMergedWithOriginal.class);
+
+            JavaPairRDD<CharSequence, DocumentText> documentTextsById = documentTexts
+                    .mapToPair(x -> new Tuple2<>(x.getId(), x));
+
+            JavaPairRDD<CharSequence, List<CharSequence>> documentMetadatasById = documentMetadatas
+                    .filter(x -> Objects.nonNull(x.getImportedAuthors()))
+                    .filter(x -> x.getImportedAuthors().stream().allMatch(y -> StringUtils.isNotBlank(y.getFullname())))
```
marekhorst: Do I get it right this `filter` operation refers to a document records not author records? If this is true, wouldn't it exclude documents having only some authors with blank names (and some having full name properly set)?

marekhorst: What I mean is: we should accept metadata records if at least one author full name was defined.

przemyslawjacewicz: Changed mapping/filtering to process authors using this schema
  - when `importedAuthors` is `null` -> map to empty list
  - when `importedAuthors` is an empty list -> forward as empty list
  - when `importedAuthors` has authors with `fullname` as `null` -> filter out authors with `null` fullname, include others

I've also added test data to cover these scenarios.

Issue #1004, opened 2019-11-20
Add proper handling for nullable fields when exporting patents


przemyslawjacewicz:
Currently patent schema allows fields not used in patent mining to be optional.  However, during exporting nullability is not properly handled e.g. no null-safe operations are used. This is not a problem currently as EPO patents file does not contain nulls for optional fields. This will become an issue when the file will start to contain nulls.

Solution is to use null-safe operations when exporting patent entities e.g. `Optional`.

przemyslawjacewicz:
Closed in #1007 .

PR #1008, opened 2019-11-22
Revert "Patents reference extraction - author flag addition"

przemyslawjacewicz:
Reverts openaire/iis#1006


przemyslawjacewicz:
Changes in #1006 do not break patents mining, so can safely be a part of #1004 and revert is not necessary.

PR #1028, opened 2019-12-18
Johnfouf project mining

johnfouf:


marekhorst:
We have incorporated those changes in #1031 PR and supplemented them with dedicated integration test case for ANR.

PR #940, opened 2019-03-06
update RPF

johnfouf:
Slightly update RPF mining rules, perhaps it'll find some links now

PR #1036, opened 2020-01-21
Update RPF

przemyslawjacewicz:
This PR is a followup after merging PR #940 into a new base branch.


marekhorst:
This PR is definitely outdated but I am trying to agree on what to do with it now in this redmine ticket comment: https://support.openaire.eu/issues/4042#note-26.

marekhorst:
We might get back to RPF mining optimization:

https://support.openaire.eu/issues/4042#note-27

so we should keep that branch and PR.

marekhorst:
This is an old PR, I just created a dedicated ticket to work on RIF/RPF:
https://github.com/openaire/iis/issues/1535

Issue #991, opened 2019-10-22
Prevent mining modules from occupying cluster resources for too long


marekhorst:
The problem is related to jobs, mostly relying on Madis, where each attempt (out of thousands of them) is locking cluster resources for a very long time (hours). This may lead other, independently executed tasks (e.g. jobs triggered by zeppelin or provisioning subsystem when deployed on IIS cluster) to wait hours before the resources occupied by IIS job are freed and made available for allocation.

The most desired solution would be to improve IIS mining modules performance but this is not always easily achievable (e.g. #951). An alternative approach would be to split data sent to Madis into more chunks resulting in creating more attempts. More attempts would mean each attempt will take less time to process the data and the resources will be returned back to pool more frequently. Having fair scheduler in place will guarantee each job will be given an equal share of resources over time.

marekhorst:
In general there are two ways of encapsulating Madis scripts within IIS:
* streaming mapreduce actions (e.g. [project reference extraction](https://github.com/openaire/iis/blob/master/iis-wf/iis-wf-referenceextraction/src/main/resources/eu/dnetlib/iis/wf/referenceextraction/project/main_sqlite/oozie_app/workflow.xml) `workflow.xml` oozie definition)
* spark job ([documents classification](https://github.com/openaire/iis/blob/master/iis-wf/iis-wf-documentsclassification/src/main/java/eu/dnetlib/iis/wf/documentsclassification/DocumentClassificationJob.java) java class)

Due to long lasting madis execution over a single chunk of data task timeout (defined as (`mapreduce.task.timeout`) had to be increased significantly reaching 8h for project reference extraction. Real attempt execution time varies from 20m up to 4h50m (3543 attempts in total).

marekhorst:
One potential spark driven solution leading to increasing number of tasks can be achieved by applying `repartition(numOfPartitions)` on an RDD provided by `loadJavaRDD()` method execution (just after loading the data). 

Notice: repartition means full reshuffle of data. Coalesce is much better performing but it can only compact the data so we cannot use it to increase the number of tasks.

marekhorst:
I had some troubles with finding working solution for a streaming approach so here is a bunch of details I have gathered while looking for viable options:
* number of splits = number of files
* `FileInputFormat` splits only files larger than HDFS block
* decreasing max split size does not seem to have an effect (more details below)
[...]

PR #1043, opened 2020-02-10
Marekhorst 991 prevent mining modules from occupying cluster resources for too long

marekhorst:
Detailed description is already available in #991 comments. 

This PR consists of 2 commits related to:
* documents classification madis module (spark action)
* PIG union transformer producing `DocumentText` input for all text mining madis modules (streaming actions)

Those two should cover all the cases where single madis task occupies cluster resources for a very long time.

`documentsclassificationNumberOfPartitions` is set to `$UNDEFINED$` by default (which means the repartitioning will NOT be applied) and is intended to be overridden in `config-default.xml` configuration file already available on IIS cluster.

I could not apply the same approach for PIG module because apparently there is no `dfs.blocksize` value which could instruct PIG to rely on cluster defaults. This is why default value is set explicitly to `128m` (cluster default) and overridden to `32m` in specific union function usage. This should result in 4x more madis tasks crunching the text input.

### DIFF
--- a/iis-wf/iis-wf-documentsclassification/src/main/java/eu/dnetlib/iis/wf/documentsclassification/DocumentClassificationJob.java
+++ b/iis-wf/iis-wf-documentsclassification/src/main/java/eu/dnetlib/iis/wf/documentsclassification/DocumentClassificationJob.java
@@ -3,6 +3,8 @@ package eu.dnetlib.iis.wf.documentsclassification;
 import com.beust.jcommander.JCommander;
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
+
+import eu.dnetlib.iis.common.WorkflowRuntimeParameters;
 import eu.dnetlib.iis.common.java.io.HdfsUtils;
 import eu.dnetlib.iis.common.schemas.ReportEntry;
 import eu.dnetlib.iis.common.spark.JavaSparkContextFactory;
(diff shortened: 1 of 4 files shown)


iis-wf/iis-wf-documentsclassification/src/main/java/eu/dnetlib/iis/wf/documentsclassification/DocumentClassificationJob.java
```diff
@@ -51,10 +53,15 @@ public static void main(String[] args) throws IOException {
 
             sc.sc().addFile(params.scriptDirPath, true);
 
-            JavaRDD<ExtractedDocumentMetadataMergedWithOriginal> documents = avroLoader
+            JavaRDD<ExtractedDocumentMetadataMergedWithOriginal> rawDocuments = avroLoader
                     .loadJavaRDD(sc, params.inputAvroPath, ExtractedDocumentMetadataMergedWithOriginal.class);
             
-            JavaRDD<DocumentMetadata> metadataRecords = documents
+            JavaRDD<ExtractedDocumentMetadataMergedWithOriginal> repartDocuments = 
+                    (StringUtils.isNotBlank(params.numberOfPartitions) && !WorkflowRuntimeParameters.UNDEFINED_NONEMPTY_VALUE.equals(params.numberOfPartitions))
```
przemyslawjacewicz: A method could be introduced to encapsulate repartitioning condition e.g. `shouldRepartition(...)`.

marekhorst:
Merged with this squashed commit: 317ae9f6a023ef684af4f2fa74cc5e5bf41efc32.

PR #1063, opened 2020-04-13
Marekhorst 973 integrate patents reference extraction rebased

johnfouf:


marekhorst:
Apparently forgot to close it. Looking at the proposed changes they are already part of the master branch.

PR #1068, opened 2020-04-21
Marekhorst 1067 update pdb database and add support for textsnippet field

marekhorst:
This merges Yannis' change being part of #1054 pull request merged with `openaire:marekhorst_1067_update_PDB_database_and_add_support_for_textsnippet_field` branch as it required to be supplemented with the set of changes on IIS side.

Those changes were required due to introducing `textsnippet` field returned by the mining script. Since we already had `DocumentToConceptId` schema version extended with `textsnippet` field defined for community mining I have decided to move it to `common` package and reuse the same schema in PDB case.


marekhorst:
> Should we commit this binary file `iis-wf/iis-wf-referenceextraction/src/main/resources/eu/dnetlib/iis/wf/referenceextraction/pdb/main/oozie_app/lib/scripts/pdb.db` ? It originates from Yannis's pull request and could be a wrongly committed result of sqllite db build.

Yes, this is one of the key and expected changes: updating PDB database.

przemyslawjacewicz:
Ok, approving.

PR #1035, opened 2020-01-21
Add INNOVIRIS

przemyslawjacewicz:
This PR is a followup after merging PR #942 into a new base branch.

PR #1034, opened 2020-01-21
Update innoviris

przemyslawjacewicz:
This PR is a followup after merging PR #944 into a new base branch.

PR #1074, opened 2020-05-04
Marekhorst 1073 integrate innoviris projects mining

marekhorst:


przemyslawjacewicz:
We have two PRs that seem to be related to this one: #1034 and #1035 . Both relate to 'innoviris' (judging by their names) so we should check if they are still valid.

przemyslawjacewicz:
Actually there are three PRs relating to 'innoviris': #1034, #1035, #1059.

marekhorst:
> Actually there are three PRs relating to 'innoviris': #1034, #1035, #1059.

Yes, those are obsolete PRs. They were issued during development phase and were not meant to be integrated with IIS master branch.

PR #1076, opened 2020-05-14
Closes #1075: Make InfoSpace metadata importer module immune to missing values in complex Field objects

### LINKED ISSUE #1075: Make InfoSpace metadata importer module immune to missing values in complex Field objects
When importing the InfoSpace graph which was produced by an updated graph builder procedure the importer process failed with `NullPointerException` due to the  `Result#getDateofacceptance()` returning `Field<String>` which is not `null` but the `value` within this object was `null`. 

This is a minor but breaking change in a contract between the graph builder and IIS importer. We should make IIS immune to such cases.

marekhorst:
This PR introduces additional checks when importing InfoSpace metadata from the graph.

### DIFF
--- a/iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtils.java
+++ b/iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtils.java
@@ -1,7 +1,5 @@
 package eu.dnetlib.iis.wf.importer.infospace.converter;
 
-import java.time.Year;
-import java.time.format.DateTimeFormatter;
 import java.util.Collection;
 import java.util.List;
 import java.util.stream.Collectors;
@@ -38,13 +36,17 @@ public abstract class MetadataConverterUtils {
     }
     
     /**
-     * Extracts year out of the date defined in yyyy-MM-dd format.
+     * Extracts year out of the date defined in yyyy-MM-dd with possible single digit month and day part.
      */
-    public static Year extractYearOrNull(String date, Logger log) {
-        try {
-            return Year.parse(date, DateTimeFormatter.ISO_LOCAL_DATE);
-        } catch (Exception e) {
-            log.warn("unsupported, non integer, format of year value: " + date);
+    public static Integer extractYearOrNull(String date, Logger log) {
+        if (StringUtils.isNotBlank(date) && date.indexOf('-') == 4) {
+            try {
+                return Integer.valueOf(date.substring(0, date.indexOf('-')));    
+            } catch (NumberFormatException e) {
+                log.warn("unsupported, non integer, format of year value: " + date);
+                return null;
+            }
+        } else {
             return null;
         }
     }
--- a/iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtils.java
+++ b/iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtils.java
@@ -2,9 +2,9 @@ package eu.dnetlib.iis.wf.importer.infospace.converter;
 
 import java.time.Year;
 import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeParseException;
 import java.util.Collection;
 import java.util.List;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
 import org.apache.log4j.Logger;
@@ -26,12 +26,14 @@ public abstract class MetadataConverterUtils {
     /**
      * Extracts values from {@link StructuredProperty} list. Checks DataInfo
      * element whether this piece of information should be approved.
+     * Does not accept null fieldApprover or source collection. Skips null values stored in this collection.
      * 
      */
     public static List<String> extractValues(Collection<StructuredProperty> source, FieldApprover fieldApprover) {
         return source.stream()
                 .filter(x -> fieldApprover.approve(x.getDataInfo()))
                 .map(StructuredProperty::getValue)
+                .filter(Objects::nonNull)
                 .collect(Collectors.toList());
     }
     
@@ -41,7 +43,7 @@ public abstract class MetadataConverterUtils {
     public static Year extractYearOrNull(String date, Logger log) {
         try {
             return Year.parse(date, DateTimeFormatter.ISO_LOCAL_DATE);
-        } catch (DateTimeParseException e) {
+        } catch (Exception e) {
             log.warn("unsupported, non integer, format of year value: " + date);
             return null;
         }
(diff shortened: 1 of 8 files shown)


iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtilsTest.java
```diff
@@ -0,0 +1,106 @@
...
+import eu.dnetlib.dhp.schema.oaf.StructuredProperty;
+import eu.dnetlib.iis.wf.importer.infospace.approver.FieldApprover;
+
+/**
+ * {@link MetadataConverterUtils} test class. 
+ * 
+ * @author mhorst
+ *
+ */
+public class MetadataConverterUtilsTest {
```
przemyslawjacewicz, line 28: IDE hint for all method with throws clause: IDE hint: Exception 'java.lang.Exception' is never thrown in the method

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtilsTest.java
```diff
@@ -0,0 +1,106 @@
...
+    @Test
+    public void testExtractValues() throws Exception {
+        // given
+        @SuppressWarnings("serial")
+        FieldApprover notNullDataInfoFieldApprover = new FieldApprover() {
+            @Override
+            public boolean approve(DataInfo dataInfo) {
+                return dataInfo != null;
+            }
+        };
```
przemyslawjacewicz, line 82: IDE hint: this can be substituted with `FieldApprover notNullDataInfoFieldApprover = Objects::nonNull;`

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtils.java
```diff
@@ -26,12 +26,14 @@ private MetadataConverterUtils() {}
...
      * Extracts values from {@link StructuredProperty} list. Checks DataInfo
      * element whether this piece of information should be approved.
+     * Does not accept null fieldApprover or source collection. Skips null values stored in this collection.
      * 
      */
     public static List<String> extractValues(Collection<StructuredProperty> source, FieldApprover fieldApprover) {
         return source.stream()
                 .filter(x -> fieldApprover.approve(x.getDataInfo()))
                 .map(StructuredProperty::getValue)
+                .filter(Objects::nonNull)
```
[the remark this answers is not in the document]
marekhorst: It won't hurt us as bad as nulls but you're right, we can drop them here as well.

I've added this case in the unit test as well.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtilsTest.java
marekhorst: Fixed all occurrences.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtilsTest.java
marekhorst: This is quite nice. 

At first glance the level of abstraction here could be overwhelming, at least until realizing functional interfaces or SAM classes can be treated in a very special way since java8.

I need to enable those neat little hints in my IDE.

marekhorst:
I've just pushed another commit with code review fixes.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtilsTest.java
przemyslawjacewicz: Yes, it is quite nice and concise. I generally start with explicit anonymous class definition and follow IDE hints to replace any code parts with more abstract versions.

przemyslawjacewicz:
Approving.

marekhorst:
I had to introduce one another commit reverting year extraction changes introduced in f814f3973d969f7aa3212d45315079d9e279094c. It turned out it discards lots of valid dates in e.g. `yyyy-M-d` format.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtils.java
```diff
@@ -38,13 +36,17 @@ private MetadataConverterUtils() {}
...
     /**
-     * Extracts year out of the date defined in yyyy-MM-dd format.
+     * Extracts year out of the date defined in yyyy-MM-dd with possible single digit month and day part.
      */
-    public static Year extractYearOrNull(String date, Logger log) {
-        try {
-            return Year.parse(date, DateTimeFormatter.ISO_LOCAL_DATE);
-        } catch (Exception e) {
-            log.warn("unsupported, non integer, format of year value: " + date);
+    public static Integer extractYearOrNull(String date, Logger log) {
```
przemyslawjacewicz, line 41: I think you could do this change two ways:
1. keep `Year` as return value and change only the parsing code part
```java
    public static Year extractYearOrNull(String date, Logger log) {
        try {
            return Year.parse(date.substring(0,  date.indexOf('-')), DateTimeFormatter.ofPattern("yyyy"));
        } catch (DateTimeParseException e) {
            log.warn("unsupported, non integer, format of year value: " + date);
            return null;
        }
    }
``` 
I think working with `Year` is better than working with integer later.
2. change the return to `Integer` and do not check for blank or index of  `-`: the only check that is relevant is the result of parsing, any bad format of `date` will eventually result in exception so it doesn't matter if we do the any checks
```java
    public static Integer extractYearOrNull(String date, Logger log) {
        try {
            return Integer.valueOf(date.substring(0, date.indexOf('-')));
        } catch (DateTimeParseException e) {
            log.warn("unsupported, non integer, format of year value: " + date);
            return null;
        }
    }
```
I'd go with option number one but the second is also ok.

marekhorst: Yes, there are pros and cons of either one. I'd go with 1st but with two additional updates:
1) dealing with strings without dashes to avoid `ArrayIndexOutOfBoundsException`
2) treating empty date cases differently than invalid ones (to avoid excessive logging).

One other case we could consider dealing with is simple four digit string but this case does not appear among the input data so maybe we should not introduce premature optimization.

I just pushed another commit.

przemyslawjacewicz:
Approving again :)

marekhorst:
Merged with IIS master branch with this squashed commit: 427c496e795aaa8930545aa314a9c88fecbe9003.

PR #1081, opened 2020-05-24
Closes #1000: Export IIS metrics to prometheus pushgateway

### LINKED ISSUE #1000: Export IIS metrics to prometheus pushgateway
Currently IIS execution report including various metrics is generated during IIS exporting phase and is made available as JSON file and via the oozie API after IIS workflow is completed.

One problem with this approach is the fact we need to wait until IIS finishes its job to know the metrics and potentially to spot anomalies. Some problems could be spotted at pretty early stage, e.g. during IIS import, so it would be reasonable to export metrics as soon as they were generated to avoid losing precious time.

We could take advantage of already existing prometheus and grafana instances to gather and present IIS metrics. Prometheus push gateway was already deployed and is ready to be used by IIS workflows. Having IIS hooked up with prometheus would expand whole new set of capabilities such as spotting regression based on an insight into the history of IIS executions, alerting, graphical presentation etc.

Since each module already covered with reporting functionality produces HDFS datastore holding [`ReportEntry`](https://github.com/openaire/iis/blob/master/iis-schemas/src/main/avro/eu/dnetlib/iis/common/ReportEntry.avdl) avro records we could quite easily provide generic action relying on those records (coming from one or more modules) and sending them to prometheus push gateway. In the first iteration we could place such action after each IIS main phase: import, processing, export. This way we could e.g. spot import related anomalies just after import phase finishes. Later on we could consider introducing more exporting checkpoints what will result in decreasing the delay between metrics generation time and their availability in prometheus.

We should agree on a naming convention of exported metrics, probably by relying on current report entry name, taking an environment into account (beta vs production, we could rely on user name triggering IIS job: either `dnet.beta` or `dnet.production`). Also report entry type (`counter` or `duration`) could be distinguished.

New solution should be implemented and defined in `iis-wf-report` module and it should be extremely fault tolerant (any error during pushing the results to pushgateway should not result in job interruption). Already implemented reporting mechanism could be still operating alongside.

przemyslawjacewicz:
This is a follow up of PR #1072 . This PR changes the names of report entry keys to allow for better prometheus metric representation, mainly
* report entries key parts reorganization to allow better usage of labels for related metrics
* changing the names of report entries to plural versions i.e. `reference -> references`

This PR also contains a fix to pushgateway metric pushing test in primary processing tests with addition of COVID-19 mining reports.


iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/processing/oozie_app/workflow.xml
```diff
@@ -2015,7 +2015,7 @@
             <arg>-Preport.processing.documentSimilarity.duration=transformers_documentssimilarity,documentssimilarity_chain</arg>
             <arg>-Preport.processing.citationMatching.direct.duration=citationmatching_direct</arg>
             <arg>-Preport.processing.citationMatching.fuzzy.duration=citationmatching_fuzzy</arg>
-            <arg>-Preport.processing.citationMatching.total.duration=citationmatching_fuzzy,citationmatching_direct,transformers_citations_from_referencemetadata,citations_union</arg>
```
marekhorst, line 2018: I think in this particular case we should not rename `total` because it refers to total amount of time required by the all citation matching phases (not to the total number of references which are deduped). We should probably drop the `total` part of the key to stay in compliance with other algos (`processing.documentSimilarity.duration`, `processing.affiliationMatching.duration`, `processing.documentClassification.duration`).

iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/output/report/primary-processing-execution-times.json
```diff
@@ -1,7 +1,7 @@
 {"key": "processing.documentClassification.duration", "type": "DURATION", "value": "$(LONG_RANGE: [1000, 170000])"}
 {"key": "processing.citationMatching.direct.duration", "type": "DURATION", "value": "$(LONG_RANGE: [1000, 150000])"}
 {"key": "processing.citationMatching.fuzzy.duration", "type": "DURATION", "value": "$(LONG_RANGE: [1000, 900000])"}
-{"key": "processing.citationMatching.total.duration", "type": "DURATION", "value": "$(LONG_RANGE: [1000, 1500000])"}
```
marekhorst, line 4: Let's drop `total` here (instead of renaming to `deduped`) as already justified in one of the comments above.

iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/output/report/pushgateway.json
```diff
@@ -1,51 +1,54 @@
...
+{"key":"processing.citationMatching.citDocReferences.deduped","type":"COUNTER","value":"2"}
 {"key":"processing.citationMatching.direct.duration","type":"DURATION","value":"$(LONG_RANGE: [1000, 150000])"}
-{"key":"processing.citationMatching.doc.direct","type":"COUNTER","value":"1"}
-{"key":"processing.citationMatching.doc.fuzzy","type":"COUNTER","value":"1"}
-{"key":"processing.citationMatching.doc.total","type":"COUNTER","value":"1"}
+{"key":"processing.citationMatching.docs.direct","type":"COUNTER","value":"1"}
+{"key":"processing.citationMatching.docs.fuzzy","type":"COUNTER","value":"1"}
+{"key":"processing.citationMatching.docs.deduped","type":"COUNTER","value":"1"}
 {"key":"processing.citationMatching.fuzzy.duration","type":"DURATION","value":"$(LONG_RANGE: [1000, 900000])"}
-{"key":"processing.citationMatching.total.duration","type":"DURATION","value":"$(LONG_RANGE: [1000, 1500000])"}
```
marekhorst, line 9: Let's drop `total` here (instead of renaming to `deduped`) as already justified in one of the comments above.

iis-wf/iis-wf-report/src/test/resources/eu/dnetlib/iis/wf/report/builder/test/input/processing_times_report.json
```diff
@@ -1,4 +1,4 @@
-{"key":"processing.citationMatching.total.duration", "type":"DURATION", "value": "22336000"}
```
marekhorst, line 1: Let's drop `total` here (instead of renaming to `deduped`) as already justified in one of the comments above.

iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/output/report/primary-processing-execution-times.json
```diff
@@ -13,5 +13,4 @@
 {"key": "processing.referenceExtraction.patent.duration", "type": "DURATION", "value": "$(LONG_RANGE: [1000, 700000])"}
 {"key": "processing.docOrgMatching.affiliationBased.duration", "type": "DURATION", "value": "$(LONG_RANGE: [1000, 200000])"}
```
marekhorst, line 14: Let me add one comment here: I have found `processing.docOrgMatching.duration` entry in the spreadsheet but I don't think we have added this metric (as it wasn't present before).

iis-wf/iis-wf-referenceextraction/src/main/resources/eu/dnetlib/iis/wf/referenceextraction/dataset/output_transformer/oozie_app/workflow.xml
```diff
@@ -75,7 +75,7 @@
             <main-class>eu.dnetlib.iis.common.java.ProcessWrapper</main-class>
             <arg>eu.dnetlib.iis.common.report.PigCountersReportGenerator</arg>
             <arg>-PpigCounters=${hadoop:counters('transformer')}</arg>
-            <arg>-Preport.processing.referenceExtraction.dataset.total.reference={dedupDocumentToDataset.RECORD_WRITTEN}</arg>
```
marekhorst, line 78: I think due to massive changes in [mapping spreadsheet](https://docs.google.com/spreadsheets/d/1q3Kx1zwW3zrSWvfeI6ePM4i3338SymOvLN86z8Xap90/edit#gid=712041498) we have overlooked this case as a candidate for `total`->`deduped` rename. I guess we could align it with `processing.citationMatching.*` and `processing.docOrgMatching.*` cases.

Especially seeing the source for this metric defined as `dedupDocumentToDataset` :).

marekhorst: One argument for this change: avoiding enquiries about `total` not being equal `datacite`+`opentrials` :) 

This would somehow let us use `total` metric name suffix ONLY for the 3 "aggregated" cases, namely: 
* `processing.referenceExtraction.community.references`
* `processing.referenceExtraction.concept.references`
* `processing.referenceExtraction.project.references`

where sum of all `byrootid.*` / `byfunder.*` metrics is in fact equal to its `total` metric.

iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/processing/oozie_app/workflow.xml
przemyslawjacewicz: Ok, dropping `total` part, should've spotted this one by looking at primary execution times because the one with `deduped` part stands out :/

iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/output/report/primary-processing-execution-times.json
przemyslawjacewicz: I think actually we're missing two reports here. We have three algos for docOrgMatching: affiliationBased, projectBased, deduped; only affiliationBased and projectBased are covered by reports. So I think we need one more report for deduped algo and one more for total docOrgMatching:
* `processing.docOrgMatching.deduped.duration`
* `processing.docOrgMatching.duration`
Am I right ?

iis-wf/iis-wf-referenceextraction/src/main/resources/eu/dnetlib/iis/wf/referenceextraction/dataset/output_transformer/oozie_app/workflow.xml
przemyslawjacewicz: Ok, renaming `total` -> `deduped`, it was a good decision to split the renaming into separate PR, because I think we could have missed things like this :)

iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/output/report/primary-processing-execution-times.json
marekhorst: Well, I'd go with `processing.docOrgMatching.duration` but skip `processing.docOrgMatching.deduped.duration` as pretty irrelevant. We don't measure deduplication for other algos as well.

przemyslawjacewicz:
Pushing recent updates, I've started integration tests on this branch as a final check for report entries key changes.

marekhorst:
OK, let's merge it (as soon as the tests are done).

przemyslawjacewicz:
Unit and integration tests are good, merging.

PR #1094, opened 2020-06-17
Closes #1085: Fine-tune spark caching in InfoSpace metadata importer module

### LINKED ISSUE #1085: Fine-tune spark caching in InfoSpace metadata importer module
While debugging #1075 problem I have discovered `DocumentMetadataConverter` code being triggered twice for a tested publication. 

This can happen because an RDD is re-evaluated again each time an action is invoked on that RDD which is the case e.g. when doing both: persisting an RDD outcome on HDFS and counting the metrics for further `ReportEntry` persistence. This is the case (persisting and metrics calculation) for almost all IIS modules. 

Our solution to minimize the need for such re-evaluation relies on performing `RDD#cache()` at appropriate execution points. This was also introduced in InfoSpace metadata importer module but:

> Spark DataFrame or Dataset cache() method by default saves it to storage level `MEMORY_AND_DISK` because recomputing the in-memory columnar representation of the underlying table is expensive. Note that this is different from the default cache level of `RDD.cache()` which is ‘MEMORY_ONLY‘.

which means we rely on `MEMORY_ONLY` storage level when manipulating on RDDs level. 

If the size of RDDs to be stored exceeds available memory then caching has no effect leading to re-evaluation of RDDs which was confirmed in `DocumentMetadataConverter`. Currently there are 130M `Result` entities in the graph and we should not even try to fit it in memory.

We should run several experiments on different storage levels and find the most appropriate one. It may also turn out the RDDs re-evaluation is the best performing solution.

marekhorst:
Changing storage level to DISK_ONLY as the most suitable one (more details in #1085 notes).

### DIFF
--- a/iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJob.java
+++ b/iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJob.java
@@ -15,6 +15,7 @@ import org.apache.spark.api.java.function.MapFunction;
 import org.apache.spark.sql.Encoder;
 import org.apache.spark.sql.Encoders;
 import org.apache.spark.sql.SparkSession;
+import org.apache.spark.storage.StorageLevel;
 
 import com.beust.jcommander.JCommander;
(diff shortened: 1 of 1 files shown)


marekhorst:
Forgot to assign the reviewer :)

PR #1105, opened 2020-07-13
Closes #1104: Increase mapreduce.task.timeout for all referenceextraction modules to 12h

### LINKED ISSUE #1104: Increase mapreduce.task.timeout for all referenceextraction modules to 12h
Due to the recent RI reference extraction job failure caused by the attempt timeout we need to increase the currently defined timeout value (4h) to the safe 12h. We should align all the other madis-based mining algos timeout just to be on a safe side.

The long lasting attempt problem was caused by two factors:
1) large number of authors (2855) causing madis to crunch single record for 3m30s
2) large number of records (96) having large number of authors in a single batch of data given to the mapper

RI algo requires text to be supplemented with meta so we end up with 96 records having 2855 authors in a single data batch. Those records require 5-6h to be processed so te 4h timeout value was way to low.

Increasing timeout value is just a quick fix but in a long run we need to change our goals when picking fulltext for mining. Up until now we've decided to accept all just to mine more and because we don't really know which plaintext is valid. So if we have two deduplicated pubs given the same dedup id we process plaintexts coming from both pubs. So the potential change to introduce within the IIS is to stop identifying plaintexts simply with OAid but to identify them by content checksum and avoid generation of duplicates. We could rely on checksums generated by ObjectStore. We would still have duplicates if the files for the deduplicated records are not identical (different versions etc) but if they were binary the same - we would end up with single plaintext entry within the IIS.

marekhorst:
The urgent need was for research-initatives but I have decided to align also all the other reference extraction modules.

### DIFF
--- a/iis-wf/iis-wf-referenceextraction/src/main/resources/eu/dnetlib/iis/wf/referenceextraction/community/main_sqlite/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-referenceextraction/src/main/resources/eu/dnetlib/iis/wf/referenceextraction/community/main_sqlite/oozie_app/workflow.xml
@@ -104,7 +104,7 @@
 					by process storing plaintexts into the database -->
 				<property>
 					<name>mapreduce.task.timeout</name>
-					<value>14400000</value>
+					<value>43200000</value>
 				</property>
                 <property>
                     <name>oozie.action.external.stats.write</name>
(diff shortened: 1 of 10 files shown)


marekhorst:
Yes, we could even make it similar to the spark related parameters which are currently the same for all the spark modules and define it using single workflow parameter.

The reason the value was defined statically within the WF definition was the assumption it should not change (as it was always defined with pretty large margin). Reality proved this was a bad assumption...

PR #1120, opened 2020-08-04
Closes #1106: Deduplicate contents by checksum

### LINKED ISSUE #1106: Deduplicate contents by checksum
Currently IIS accepts multiple contents for each publication in order to have as many plaintexts to mine from as possible. Those multiple contents may come as a result of deduplication process where two or more separate entities are considered as one and all associated contents are being linked to a new representative.

While debugging recent research-initiative reference extraction timeout failure it turned out we've had 96 text entries associated with the same publication (deduped) identifier. When joining those 96 records with very large number of authors (2855) which happened to be associated with this particular publication we've generated 96 "monster" records. Limiting large number of authors will be addressed in a separate task (#1107) but we could also do much better while dealing with multiple plaintexts.

One simple approach would be to spot all the duplicates for given deduped publication by comparing content checksum provided by ObjectStore. This way we could eliminate at least some of the content duplicates at the importing phase. So this solution would be applied to each publication separately.

The other approach could be understood as infospace-wide: we could identify all the plaintexts within the IIS by checksum. Currently they are identified by OpenAIRE identifier but we could rely on checksums instead to avoid streaming down to the text mining modules the same text twice. Each text will be processed once and the results will be propagated to all the OpenAIRE entities having given content among associates. This would require introducing major changes in IIS internal model.

Both approaches could be combined.

marekhorst:
Deduplication is done by grouping records by id and content checksum and is performed by the dedicated PIG script defined as one of the `content_url/chain` phases (after identifiers mapping and filtering, before dispatching by mime-type). 

An alternative solution was to either make it a part of already defined phases or reimplement whole `content_url/chain` submodules in spark. None of already existing modules was a good candidate to incorporate deduplication into (due to a pretty well defined expectations against those modules) while reimplementing whole workflow as a spark job would take a massive amount of time without any significant performance-gain. Whole `content_url/chain` chain takes 26 mins to process the data and the most time consuming subpart is ObjectStore metadata import (19 mins) while the transformations which could be a subject for refactoring take 7 minutes.

### DIFF
--- a/iis-wf/iis-wf-import/src/test/resources/eu/dnetlib/iis/wf/importer/content_url/dedup/sampletest/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-import/src/test/resources/eu/dnetlib/iis/wf/importer/content_url/dedup/sampletest/oozie_app/workflow.xml
@@ -1,0 +1,78 @@
+<?xml version="1.0"?>
+<workflow-app xmlns="uri:oozie:workflow:0.4" name="test-importer_content_url_dedup_sampletest">
+
+    <global>
+        <job-tracker>${jobTracker}</job-tracker>
+        <name-node>${nameNode}</name-node>
+        <configuration>
+            <property>
+                <name>mapreduce.job.queuename</name>
+                <value>${queueName}</value>
+            </property>
+            <property>
+                <name>oozie.launcher.mapred.job.queue.name</name>
+                <value>${oozieLauncherQueueName}</value>
+            </property>
+        </configuration>
+    </global>
+    
+	<start to="producer"/>
+    
+    <action name="producer">
+        <java>
+            <prepare>
+                <delete path="${nameNode}${workingDir}/producer" />
+                <mkdir path="${nameNode}${workingDir}/producer" />
+            </prepare>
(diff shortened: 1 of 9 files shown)


iis-wf/iis-wf-import/src/test/resources/eu/dnetlib/iis/wf/importer/content_url/dedup/sampletest/oozie_app/workflow.xml
```diff
@@ -0,0 +1,78 @@
...
+        <error to="fail" />
+    </action>
+
+	<action name="consumer">
+        <java>
+            <!-- This is simple wrapper for the Java code -->
+            <main-class>eu.dnetlib.iis.common.java.ProcessWrapper</main-class>
+            <!-- The business Java code that gets to be executed -->
+            <arg>eu.dnetlib.iis.common.java.jsonworkflownodes.TestingConsumer</arg>
+            <!-- All input and output ports have to be bound to paths in HDFS -->
```
przemyslawjacewicz, line 63: Comments are not necessary here, we've already have them in iis-examples.

marekhorst:
Closed with this commit after rebase: https://github.com/openaire/iis/commit/d53e8ae42494de7f28bc6a2548b5e3096bef2ce5.

PR #1123, opened 2020-08-05
Closes #1122: SoftwareHeritage importer fails with NoSuchFieldError:INSTANCE

### LINKED ISSUE #1122: SoftwareHeritage importer fails with NoSuchFieldError: INSTANCE
When trying to sync SH database `SoftwareHeritageOriginsImporter` surprisingly failed with:

```
aused by: java.lang.NoSuchFieldError: INSTANCE
	at org.apache.http.conn.ssl.SSLConnectionSocketFactory.<clinit>(SSLConnectionSocketFactory.java:151)
	at org.apache.http.impl.client.HttpClientBuilder.build(HttpClientBuilder.java:977)
	at eu.dnetlib.iis.wf.importer.software.origins.SoftwareHeritageOriginsImporter.buildHttpClient(SoftwareHeritageOriginsImporter.java:210)
	at eu.dnetlib.iis.wf.importer.software.origins.SoftwareHeritageOriginsImporter.run(SoftwareHeritageOriginsImporter.java:121)
	at eu.dnetlib.iis.common.java.ProcessWrapper.run(ProcessWrapper.java:49)
```

~After inspecting the changes in our code it looks like the problem was induced by introducing explicit `httpclient` dependency in main `pom.xml` file as a part of #1099 solution. Seems to be the clash between the different versions of http libraries.~
~If this dependency could be dropped as a part of retracting the TARA caching solution in #1117 then we would probably automatically solve this issue. Otherwise we would need to dig deeper and align the http libraries versions.~

The `httpclient` explicit dependency was introduced as a part of #1070 major patent refactoring where HTTP communication with EPO endpoint was refactored and SH communication followed the same refactoring. Apparently the communication with SH endpoint is handled differently leading to the mentioned issue.

marekhorst:
Giving user jars precedence over the parcels bringing old version of httpclient.

### DIFF
--- a/iis-wf/iis-wf-import/src/main/resources/eu/dnetlib/iis/wf/importer/software_origins/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-import/src/main/resources/eu/dnetlib/iis/wf/importer/software_origins/oozie_app/workflow.xml
@@ -126,6 +126,12 @@
 				<delete path="${nameNode}${workingDir}/imported" />
                 <delete path="${nameNode}${output}" />
 			</prepare>
+			<configuration>
+				<property>
+				    <name>oozie.launcher.mapreduce.user.classpath.first</name>
+				    <value>true</value>
+				</property>
+			</configuration>
 			<main-class>eu.dnetlib.iis.common.java.ProcessWrapper</main-class>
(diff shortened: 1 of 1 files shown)


marekhorst:
Closed with this commit after rebase: https://github.com/openaire/iis/commit/accc50c97f569179e9c34e761942fc66ae372681.

PR #1125, opened 2020-08-07
Closes #1124: Follow SoftwareHeritage API changes: origin_from renamed to page_token

### LINKED ISSUE #1124: Follow SoftwareHeritage API changes: origin_from renamed to page_token
When issuing queries against the SH API:
https://archive.softwareheritage.org/api/1/origins/?origin_count=100&origin_from=101

the following response (code=400) is generated:

```
{
    "exception": "BadInputExc",
    "reason": "Please use the Link header to browse through result"
}
```
It turned out the `origin_from` parameter was renamed to `page_token` so now the valid link is:
https://archive.softwareheritage.org/api/1/origins/?origin_count=100&page_token=101

Before we apply simple fix on our side we should make sure we can carry on with incremental indexing relying on `page_token` as an index/position of the new starting point in an ordered sequence of records. Up until now we've been persisting the position of the most recently retrieved record to reuse it later on when picking up another round of incremental cache update.

marekhorst:
This is a simple rename of parameters: `origin_from` -> `page_token`. Should be merged with #1122 feature branch first and then both could be merged with IIS master branch.


marekhorst:
Merged with this commit https://github.com/openaire/iis/commit/e5bf3ec11bb7b74b95a183b68d1d5afd5bf6a201 after rebase.

PR #1126, opened 2020-08-10
Closes #1117: Replace TARA caching with storing TARA projects in a dedicated sqlite db instance

### LINKED ISSUE #1117: Replace TARA caching with storing TARA projects in a dedicated sqlite db instance
Long story short: already introduced TARA caching (#1099) takes more time than running TARA mining on TARA projects stored in a dedicated sqlite database instance.

TARA mining ([taraextract.sql](https://github.com/openaire/iis/blob/4323a80c0783b9598324742b3c87991cf0425970/iis-wf/iis-wf-referenceextraction/src/main/resources/eu/dnetlib/iis/wf/referenceextraction/project/tara/oozie_app/lib/scripts/taraextract.sql)), which is defined outside generic projects mining ([projects.sql](https://github.com/openaire/iis/blob/eb3f4b93c398ceba62162148075aa4b281cfceda/iis-wf/iis-wf-referenceextraction/src/main/resources/eu/dnetlib/iis/wf/referenceextraction/project/main_sqlite/oozie_app/lib/scripts/projects.sql)), is currently relying on the same SQLite database the generic projects mining does.

After running TARA caching tests on a full set of projects excluding TARA projects it turned out the algorithm still needed ~11h to make a "dry-run" over a full set of publications. This was quite surprising because the assumptions we were relying on were:

1. running projects mining for different funders on a subset of projects should not improve the performance significantly
2. TARA mining logic relies on a quite advanced NLP what may put an impact on performance

while apparently the large number of projects has significant impact on TARA mining performance. It was confirmed by a subsequent runs including 5 TARA projects only where TARA mining took 25 minutes using spark version (including input transformation) and 17 minutes using an old M/R based TARA mining.

marekhorst:
Altering `wf/referenceextraction/project/input_filter` subworkflow to produce an output with rejected records. To be used for TARA projects dispatching purposes.


marekhorst:
Hmm, I don't know, I think I would leave it as it is. 

The main reason is this filter module operates on projects only (incompatible with other types) so I don't think it is ambiguous. Besides it produced simple `output` before so I don't think we should be more verbose when adding rejected/approved.

przemyslawjacewicz:
Ok, I simply like long informative names ;-) Approving.

Issue #1096, opened 2020-06-17
Introduce caching for patent metadata retrieval


marekhorst:
Once we are done with patents mining refactoring in #1070 and introducing caching for TARA (described briefly in #1080) we should take advantage of newly refactored caching solution and cover patent metadata retrieval phase (`PatentMetadataRetrieverJob`) with caching.

This way we could save significant amount of time and limit the communication with external EPO endpoint only to the newly matched patents. EPO endpoint imposes quite severe rate limits requiring the IIS to limit the number of concurrently run exeuctors to `1` and throttle the rate of requests being send if the endpoint is overloaded.

marekhorst:
When introducing caching solution we still need to limit the number of tasks communicating with EPO endpoint concurrently. The main reason is severe rate-limiting when accessing EPO resources in parallel.

Several tests (on an already filled cache) were run in different setup to limit the number of concurrent calls:
* 1 executor with 1 core defined as `spark.dynamicAllocation.maxExecutors` and `executor-cores` spark parameters ([spark history](http://iis-cdh5-test-gw.ocean.icm.edu.pl:18089/history/application_1595483222400_4113/1/jobs/))
* 1 partition created by issuing `coalesce(1)` on imported patents ([spark history](http://iis-cdh5-test-gw.ocean.icm.edu.pl:18089/history/application_1595483222400_4308/1/jobs/))
* 1 partition created by issuing `coalesce(1)` on an RDD being transformed with map operation issuing EPO requests ([spark history](http://iis-cdh5-test-gw.ocean.icm.edu.pl:18089/history/application_1595483222400_4336/1/jobs/))

It turned out the last option was the optimal one. Spark parameters based limitation was resulting in a vast amount on jobs in tens of phases handed by a single executor what resulted in 13 minutes of execution time. 2nd and 3rd options were much better performing (5 minutes) and resulted in a limited number of jobs/phases. 3rd option seem to be better because the limitation is defined only in the place where the real communication with EPO endpoint is being handled so potentially, if the execution plan allows it, other phases involving different kinds of transformations can be run without the limitations.

PR #1135, opened 2020-08-12
Closes #1131: Change the way software webcrawler parallelism level is capped

### LINKED ISSUE #1131: Change the way software webcrawler parallelism level is capped
The number of tasks responsible for retrieving HTML pages from software repositories has to be capped due to a rate limiting imposed by the repositories.

Currently it is done by setting `spark.dynamicAllocation.maxExecutors` to `4` what guarantees no more than 4 executors will handle the caching tasks (defined along with `executor-cores=1` to make sure 4 independent operations can run in parallel, at most). The problem is this solution is quite invasive and restricts all the tasks being run as a part of `referenceextraction_softwareurl_webcrawl` spark job, not only the communication with external hosts.

The improved solution was introduced in patents caching and was described in this note: https://github.com/openaire/iis/issues/1096#issuecomment-668113876. The problem with `coalesce(numPartitions)` is it affects all the other upstream transformations within the stage but we could issue either `coalesce(numPartitions, true)` or `repartition(numPartitions)` to make sure only the communication with external hosts will be capped.

More details on limiting the number of partitions can be found in this comment: https://github.com/openaire/iis/issues/1103#issuecomment-671938466.

marekhorst:
Introducing `webcrawlNumberOfPartitionsForCrawling` parameter, set to 4 by default, describing number of partitions to be created for web pages retrieval phase.

This PR relies on a solution already introduced in patents retrieval caching (#1096).

### DIFF
--- a/iis-wf/iis-wf-referenceextraction/src/main/java/eu/dnetlib/iis/wf/referenceextraction/softwareurl/ContentRetrieverContext.java
+++ b/iis-wf/iis-wf-referenceextraction/src/main/java/eu/dnetlib/iis/wf/referenceextraction/softwareurl/ContentRetrieverContext.java
@@ -25,12 +25,15 @@ public class ContentRetrieverContext implements Serializable {
     private int maxPageContentLength;
     
     private int numberOfEmittedFiles;
-
+    
+    private int numberOfPartitionsForCrawling;
 
     public ContentRetrieverContext() {}
     
-    public ContentRetrieverContext(String contentRetrieverClassName,
-            int connectionTimeout, int readTimeout, int maxPageContentLength, int numberOfEmittedFiles) throws InstantiationException, IllegalAccessException, IllegalArgumentException, InvocationTargetException, NoSuchMethodException, SecurityException, ClassNotFoundException {
+    public ContentRetrieverContext(String contentRetrieverClassName, int connectionTimeout, int readTimeout,
+            int maxPageContentLength, int numberOfEmittedFiles, int numberOfPartitionsForCrawling)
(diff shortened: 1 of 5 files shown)


iis-wf/iis-wf-referenceextraction/src/main/java/eu/dnetlib/iis/wf/referenceextraction/softwareurl/ContentRetrieverContext.java
```diff
@@ -25,44 +25,49 @@
...
+    private int numberOfPartitionsForCrawling;
 
     public ContentRetrieverContext() {}
     
-    public ContentRetrieverContext(String contentRetrieverClassName,
-            int connectionTimeout, int readTimeout, int maxPageContentLength, int numberOfEmittedFiles) throws InstantiationException, IllegalAccessException, IllegalArgumentException, InvocationTargetException, NoSuchMethodException, SecurityException, ClassNotFoundException {
+    public ContentRetrieverContext(String contentRetrieverClassName, int connectionTimeout, int readTimeout,
+            int maxPageContentLength, int numberOfEmittedFiles, int numberOfPartitionsForCrawling)
+            throws InstantiationException, IllegalAccessException, IllegalArgumentException, InvocationTargetException,
+            NoSuchMethodException, SecurityException, ClassNotFoundException {
```
przemyslawjacewicz, line 36: Use `throws Exception` instead of 7 exceptions in `throws`.

marekhorst:
Merged with this squashed commit: https://github.com/openaire/iis/commit/cf2c99dbaa8bd967ea1b8b548c4ad312772ab69a.

PR #1144, opened 2020-09-08
Improve the "textsnippet"-field in Canadian mining.

LSmyrnaios:
This update makes more clear what the match is, as there might be multiple funders in the same textsnippet.


marekhorst:
The PR was merged with https://github.com/openaire/iis/commit/3181f4953b682ef380ed47a203f83aca462e7800 commit.

PR #1145, opened 2020-09-09
Closes #1107: Eliminate outliers by applying thresholds to exploding dimensions of InfoSpace entities

### LINKED ISSUE #1107: Eliminate outliers by applying thresholds to exploding dimensions of InfoSpace entities
This is one another outcome of debugging recent research-initiative reference extraction timeout failure (other reported issue: #1106). 

It turned out the failure mentioned above was induced by a publication having very large number of authors (2855). We could definitely limit the number of authors streamed down to mining and possibly other "exploding" dimensions of InformationSpace graph.

Similar task was already undertaken in the context of building SOLR index records where the following hard limits were introduced:
```
int MAX_EXTERNAL_ENTITIES = 50; // maximum number elements in result.externalReference[]
int MAX_AUTHORS = 200;  // maximum number elements in result.author[]
int MAX_AUTHOR_FULLNAME_LENGTH = 1000; // maximum length allowed for authors fullname
int MAX_TITLE_LENGTH = 5000; // maximum length allowed for titles
int MAX_TITLES = 10; // maximum number of elements in result.title[]
int MAX_ABSTRACT_LENGTH = 100000; // maximum length allowed for result.description (abstract)
```
The requirements against searching functionality are slightly different than against mining but this definitely indicates the dimensions we should take a closer look at.

This way we could minimize the impact of outliers on IIS performance. Research-initiative mining was already revised and it turned out the single attempt taking significantly more time to execute than the others (6h for standing out attempt vs 2h and below for others) was caused by the 2855 of authors. There could be a similar reason for 2 attempts standing out for projects mining (4h40m vs 1h and less for others) and datacite mining (3 attempts above 2h while most of the others run for 1h and below). The data chunk being processed by those standing out attempts should be analyzed in the same way it was done for RI algo.

przemyslawjacewicz:
This PR adds the support for eliminating outliers in document metadata. The central point of the implementation is infospace import job `eu.dnetlib.iis.wf.importer.infospace.ImportInformationSpaceJob` and document metadata truncation support in `eu.dnetlib.iis.wf.importer.infospace.truncator.DocumentMetadataAvroTruncator`. For convenient testing additional changes allow to use comprehensible values of truncation parameters.


iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/truncator/DocumentMetadataAvroTruncator.java
```diff
@@ -0,0 +1,200 @@
...
+        public void setStringTruncator(Function<CharSequence, CharSequence> stringTruncator) {
+            this.stringTruncator = stringTruncator;
+        }
+
+        CharSequence truncate(CharSequence x) {
+            return stringTruncator.apply(x);
+        }
+    }
+
+    public static class TitleTruncator implements Serializable {
```
marekhorst, line 64: Do we really need a dedicated Truncator instance for each field? `AbstractTruncator` looks exactly the same as the `TitleTruncator`. 

I am not sure we really need to propagate the truncators model deep down below the `DocumentMetadataAvroTruncator` for each field. I know there are the test classes for each case, which is nice, but maybe we could live with just a truncation test for `DocumentMetadata` and test truncation of all the dimensions. There are only 6 of them. I'd just apply whole truncation logic within the:
```
public DocumentMetadata truncate(DocumentMetadata source)
```
method and drop all the internal truncators. To me it feels overcomplicated.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/truncator/factory/TestDocumentMetadataAvroTruncatorFactory.java
```diff
@@ -0,0 +1,17 @@
+package eu.dnetlib.iis.wf.importer.infospace.truncator.factory;
+
+import eu.dnetlib.iis.wf.importer.infospace.truncator.DocumentMetadataAvroTruncator;
+
+public class TestDocumentMetadataAvroTruncatorFactory implements DocumentMetadataAvroTruncatorFactory {
```
marekhorst, line 5: OK, so the truncators solution, apart from being a subject for modular testing, is kind of a wrapper encapsulating the thresholds for 6 dimensions, right? 

So instead of overwriting 6 dimensions for the test we can switch the truncator implementation. It's nice, on one hand because we can keep the workflow definition compact, but on the other hand when having 6 overridable threshold parameters we could make them easily overridable when it turns out our pick statically defined in `ImportInformationSpaceConstants` was not accurate. It might be a matter of a personal opinion but I would prefer to have more flexibility in this regard.

marekhorst (changes requested):
Truncators model is quite interesting but imho quite over-complicated for such a simple truncation operation. I wouldn't even mind truncating the strings/collections directly within the `DocumentMetadataConverter` module, but both aproaches have their pros and cons.

I would leave some of its pieces but at the same time I would change the granularity of `Truncator` usage.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/truncator/DocumentMetadataAvroTruncator.java
```diff
@@ -0,0 +1,200 @@
...
+
+        public Builder setMaxKeywordLength(Integer maxKeywordLength) {
+            this.maxKeywordLength = maxKeywordLength;
+            return this;
+        }
+
+        public DocumentMetadataAvroTruncator build() {
+            DocumentMetadataAvroTruncator instance = new DocumentMetadataAvroTruncator();
+
+            if (Objects.nonNull(maxAbstractLength)) {
```
marekhorst, line 181: If the object is currently instantiated by relying on the statically defined values within `ImportInformationSpaceConstants` class do we need to check the nullability of those fields here and in similar places below?

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/truncator/factory/DefaultDocumentMetadataAvroTruncatorFactory.java
```diff
@@ -0,0 +1,13 @@
+package eu.dnetlib.iis.wf.importer.infospace.truncator.factory;
+
+import eu.dnetlib.iis.wf.importer.infospace.truncator.DocumentMetadataAvroTruncator;
+
+/**
+ * Factory for {@link DocumentMetadataAvroTruncator} providing a default implementation with production values of truncation parameters.
+ */
+public class DefaultDocumentMetadataAvroTruncatorFactory implements DocumentMetadataAvroTruncatorFactory {
```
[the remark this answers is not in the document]
przemyslawjacewicz: After changing the way of passing truncation parameters to the importer job this class is no longer needed and I've removed it.

iis-wf/iis-wf-import/src/main/resources/eu/dnetlib/iis/wf/importer/infospace/oozie_app/workflow.xml
```diff
@@ -168,7 +173,8 @@
             <arg>-outputNameDedupMapping=${output_name_dedup_mapping}</arg>
             <arg>-outputNameOrganization=${output_name_organization}</arg>
             <arg>-outputNameProjectOrganization=${output_name_project_organization}</arg>
-            
+            <arg>-documentMetadataAvroTruncatorFactoryClassName=${document_metadata_avro_truncator_factory_class_name}</arg>
```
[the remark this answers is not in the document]
przemyslawjacewicz: After changing the way of passing truncation parameters to importer job this parameter is no longer needed. Instead I've added dynamic parameters with truncation parameters. The values previously set in java code are now default values and are passed to the job with `-D` prefix.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceConstants.java
```diff
@@ -0,0 +1,13 @@
+package eu.dnetlib.iis.wf.importer.infospace;
+
+/**
+ * {@link eu.dnetlib.iis.importer.schemas.DocumentMetadata} truncation constants.
+ */
+public class ImportInformationSpaceConstants {
```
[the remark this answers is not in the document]
przemyslawjacewicz: Done, should've done that at the beginning. This change also results with the following changes:
* `DocumentMetadataAvroTruncator` and `AuthorAvroTruncator` simplification - no nested classes
*  removal of `DocumentMetadataAvroTruncatorFactory` with related classes
* unit and integration test changes

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJob.java
```diff
@@ -186,16 +192,33 @@ public static void main(String[] args) throws Exception {
...
     /**
-     * Parses given set of RDDs conveying various {@link Result} entities into a single RDD with {@link DocumentMetadata} records.
+     * Creates a factory of {@link DocumentMetadataAvroTruncatorFactory} from class name.
+     *
+     * @param className Canonical name of a class implementing {@link DocumentMetadataAvroTruncatorFactory}.
+     * @return Instance of the class.
+     */
+    private static DocumentMetadataAvroTruncatorFactory documentMetadataAvroTruncatorFactory(String className) {
+        try {
+            return (DocumentMetadataAvroTruncatorFactory) Class.forName(className).getConstructor().newInstance();
```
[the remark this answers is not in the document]
przemyslawjacewicz: Done, truncation parameters are now dynamic parameters.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/truncator/DocumentMetadataAvroTruncator.java
przemyslawjacewicz: Done, I've removed dedicated truncator classes for each field. The only reason for their implementation was to make unit testing cleaner.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/truncator/factory/TestDocumentMetadataAvroTruncatorFactory.java
przemyslawjacewicz: After changing the way of passing truncation parameters to importer job this class is no longer needed. Truncation parameters are now passed as dynamic parameters with default values. For tests the values are explicitly provided and there is no need for any 'factory' classes.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/truncator/DocumentMetadataAvroTruncator.java
przemyslawjacewicz: With this implementation the check is necessary. `maxAbstractLength` in this null-check is an `Integer' field within builder nested class. If this field was set using `setMaxAbstractLength()` method then we use it when creating abstract truncator, if this field was not set and was null then we fallback to default abstract truncator with default value created by `DocumentMetadataAvroTruncator` class.

However, builder logic is now changed as a result of changing the way truncation parameters are passed to importer job and this null-check is not needed.  `DocumentMetadataAvroTruncator` is created with explicit parameters provided in the workflow.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/truncator/StringTruncator.java
```diff
@@ -0,0 +1,24 @@
...
+     * @return Truncated string.
+     */
+    public static String truncateWithoutWordSplit(String source, int length) {
+        if (source.length() <= length) {
+            return source;
+        }
+
+        BreakIterator atWordInstance = BreakIterator.getWordInstance();
+        atWordInstance.setText(source);
+        return source.substring(0, atWordInstance.preceding(length)).trim();
```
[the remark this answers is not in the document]
przemyslawjacewicz: Discussed offline: string truncation without word split can potentially remove a large part of source string when the string is short and has a longer word at the truncation index. To avoid this kind of situations we agreed on keeping the word at truncation index. With `BreakIterator` this means using `.following` method and this change is applied here. This will also result in changes in test resources with truncated document metadata.

przemyslawjacewicz:
Finished applying changes - moving truncation parameters to workflow params was the most significant change and caused some additional changes.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/truncator/DocumentMetadataAvroTruncatorTest.java
```diff
@@ -0,0 +1,156 @@
...
+        // when
+        DocumentMetadata result = documentMetadataAvroTruncator.truncate(source);
+
+        // then
+        assertEquals(source, result);
+        verify(authorsTruncator, never()).apply(any());
+    }
+
+    @Test
+    public void shouldTruncateAuthorsWhenItIsNotNull() {
```
marekhorst, line 106: I guess we could also add a test verifying whether the number of authors is properly truncated.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/truncator/DocumentMetadataAvroTruncatorTest.java
```diff
@@ -0,0 +1,156 @@
...
+        // when
+        DocumentMetadata result = documentMetadataAvroTruncator.truncate(source);
+
+        // then
+        assertEquals(source, result);
+        verify(keywordsTruncator, never()).apply(any());
+    }
+
+    @Test
+    public void shouldTruncateKeywordsWhenItIsNotNull() {
```
marekhorst, line 135: Same here: I guess we could also add a test verifying whether the number of keywords is properly truncated.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJob.java
```diff
@@ -416,6 +457,9 @@ private static String outputPathFor(String path, String subPath) {
         
         @Parameter(names = "-outputNameProjectOrganization", required = true)
         private String outputNameProjectOrganization;
+
+        @DynamicParameter(names = "-D", description = "dynamic parameters related to document metadata truncation")
+        private Map<String, String> documentMetadataTruncationParams = new HashMap<>();
```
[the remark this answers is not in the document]
przemyslawjacewicz: The parameters are definitely not dynamic, it would be clearer to define them as normal parameters. I thought that the reason to use dynamic parameters was to keep `ImportInformationSpaceJobParameters` small and clean, but if we can accept a few more fields in this class then I'd go with adding explicit parameters. 

Looking a bit in the future if we find that we need to add even more truncation parameters I'd introduce a dedicated workflow step that would be responsible for truncation. This step would take all the needed parameters as explicit parameters to the job.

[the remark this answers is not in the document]
przemyslawjacewicz: Or maybe you think we should add a dedicated workflow step now ?

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/truncator/DocumentMetadataAvroTruncatorTest.java
przemyslawjacewicz: This unit test checks if authors truncation logic is triggered. It uses a stubbed mock for `Function<List<Author>, List<Author>>` that is responsible for truncation of authors field in `DocumentMetadataAvroTruncator`. So this test method simply checks if the result of truncation is the same as the value used for stubbing. The actual size of the list is not important, that's why I've used a `Collections.singletonList(mock(Author.class))`. The 'real' truncation mechanism is checked in `ImportInformationSpaceJobTest` and in workflow integration test.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/truncator/DocumentMetadataAvroTruncatorTest.java
przemyslawjacewicz: And the same answer applies here :)

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJob.java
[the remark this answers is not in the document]
marekhorst: I think we don't need to go that far, I mean I don't expect much more truncation-related parameters to pop-up.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/truncator/DocumentMetadataAvroTruncatorTest.java
marekhorst: OK then. I thought we are missing the unit test proving maxAuthorsSize limitation works as expected.

przemyslawjacewicz:
> Apart from three rather minor remarks I have one major one which does not seem to be addressed:
> 
> > I am not sure we should bound the truncator with any specific data type because DocumentMetadata is not the only metadata type with fields to be truncated: there is also DataSetReference. Alternatively we would need to introduce dataset truncator handling DataSetReference records.
> 
> Apart from `DocumentMetadata` we also have `DataSetReference` as a candidate for truncation because it holds potentially exploding dimensions:
> 
>     * `creatorNames` (by `max_authors_size` and `max_author_fullname_length`)
> 
>     * `titles` (by `max_titles_size` which should be added and set to `10` by default and and `max_title_length`)
> 
>     * `description` (by `max_description_length`, we could rename already existing `max_abstract_length` and reference same dimension in both places, both are relying on `Result#description` field so the `description` part of the name suits pretty well)
> 
> 
> We've lost it from the radar but I guess we should cover those objects with truncation as well.

I've just added truncation of `DataSetReference` using the description given above. To make truncation a little more generic I've added a method in `ImportInformationSpaceJob` for truncation of any avro type. Apart from that the changes with adding `DataSetReference` are minor.

marekhorst (approved):
Thanks, LGTM. 

Besides: it turns out we ended up with 1200 added lines, so in the end it wasn't that simple task we initially thought it would be. Probably most of it was related to unit and integration tests but this falls into the bright side ;)

przemyslawjacewicz:
Yes, I was also a bit surprised with the overall size of this PR !

PR #1147, opened 2020-09-14
Closes #1134: Fix multibranch build failing when more than one build is being run in parallel

### LINKED ISSUE #1134: Fix multibranch build failing when more than one build is being run in parallel
I just noticed the multibranch build for a newly created branch `marekhorst_1132_fix_http_connection_pool_handling` has failed:

http://ci.ceon.pl/job/iis_build_multibranch/job/marekhorst_1132_fix_http_connection_pool_handling/2/

with the following unit test error:

```
Path does not exist: file:/srv/yadda-jenkins/app/workspace/tp_connection_pool_handling-7BTP7MM64PLLYOZDVA67NVVWYNKYYC65ZRPJFHBI7E5YN4CASU6A%402/iis-wf/iis-wf-import/target/test-classes/eu/dnetlib/iis/wf/importer/patent/sampletest/oozie_app/input/patents_epo.tsv;
```

I've seen similar errors before but just did care too much, I thought those are some problems of an intermittent kind.

It looks like the path contains the `%402` character which might be the cause of the problem (the test run locally succeeds).

When I looked at the set of all the jobs being run for the branch:
http://ci.ceon.pl/job/iis_build_multibranch/job/marekhorst_1132_fix_http_connection_pool_handling/
I found two of them, triggered almost at the same time. I guess the first one was triggered after the branch was pushed to the origin repository while the second when issuing a pull request. @przemyslawjacewicz, could it be the case?

Anyway, if the two executions were expected then we are left with one issue to be solved only:)

I have found one resource describing very similar issue:
https://stackoverflow.com/questions/39553523/jenkins-failures-due-to-an-extra-402-character-in-the-location-address
which might be helpful in solving our problem.

przemyslawjacewicz:
This PR fixes the problem with parallel executions of jenkins builds. 

Each test resource path retrieved using classloader -> resource -> file pipeline is UTF-8 encoded and this PR adds a decode step, converting the path to a proper string pointing to resources for jenkins build. This PR adds an utility class with a method for decoding UTF-8 encoded strings and applies this method anywhere a test resource path is retrieved.


mpol:
A general doubt about the approach: https://docs.oracle.com/javase/8/docs/api/java/net/URL.html says

>  Note, the URI class does perform escaping of its component fields in certain circumstances. The recommended way to manage the encoding and decoding of URLs is to use URI, and to convert between these two classes using toURI() and URI.toURL().
>
> The URLEncoder and URLDecoder classes can also be used, but only for HTML form encoding, which is not the same as the encoding scheme defined in RFC2396.

This may not actually matter in this case, but wouldn't going the recommended route via `URI` be clearer?

mpol:
Also, independently of the approach, I think extracting the whole `(ClassLoader, String) => String` function would be more readable.

There is quite a lot happening in both `StringUtils.decodeFromUtf8(Objects.requireNonNull(cl.getResource(INPUT_PATENT_PATH)).getFile()` and `Objects.requireNonNull(cl.getResource(INPUT_PATENT_PATH)).toURI().getPath()`

iis-common/src/main/java/eu/dnetlib/iis/common/string/StringUtils.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+public class StringUtils {
+
+    /**
+     * Decodes an UTF-8 encoded string.
+     *
+     * @param source UTF-8 encoded string.
+     * @return Decoded string.
+     */
+    public static String decodeFromUtf8(String source) {
```
mpol, line 18: Not a fan of this function's name. It's actually [URI  percent-encoding](https://tools.ietf.org/html/rfc3986#section-2.1) that we are dealing with here.

marekhorst: It caught my eye as well. Maybe `decodeURL()` would be better here? Or something else describing the context more precisely.

marekhorst (changes requested):
What is so special about the patents and covid19 algos those are the ones mostly affected? Speaking other words: does it mean we don't need to fix other `JsonAvroTestUtils.readJsonDataStore()` references which are plenty and can be found in multiple mining modules?

iis-common/src/main/java/eu/dnetlib/iis/common/string/StringUtils.java
przemyslawjacewicz: Yes, the name was chosen without some more thought. I've removed the whole class and added a resource handling class `eu.dnetlib.iis.common.utils.ResourceUtils` with a dedicated method `resourcePath(ClassLoader cl, String name)` for extracting resource path with proper formatting.

przemyslawjacewicz:
> A general doubt about the approach: https://docs.oracle.com/javase/8/docs/api/java/net/URL.html says
> 
> > Note, the URI class does perform escaping of its component fields in certain circumstances. The recommended way to manage the encoding and decoding of URLs is to use URI, and to convert between these two classes using toURI() and URI.toURL().
> > The URLEncoder and URLDecoder classes can also be used, but only for HTML form encoding, which is not the same as the encoding scheme defined in RFC2396.
> 
> This may not actually matter in this case, but wouldn't going the recommended route via `URI` be clearer?

> Also, independently of the approach, I think extracting the whole `(ClassLoader, String) => String` function would be more readable.
> 
> There is quite a lot happening in both `StringUtils.decodeFromUtf8(Objects.requireNonNull(cl.getResource(INPUT_PATENT_PATH)).getFile()` and `Objects.requireNonNull(cl.getResource(INPUT_PATENT_PATH)).toURI().getPath()`

I've removed string utils class with `decodeFromUtf8` method and added a new class specifically for handling resources `eu.dnetlib.iis.common.utils.ResourceUtils`. This class has `resourcePath(ClassLoader cl, String name)` method which retrieves resource path with proper handling of special characters. It uses the `URI` approach.


> What is so special about the patents and covid19 algos those are the ones mostly affected? Speaking other words: does it mean we don't need to fix other `JsonAvroTestUtils.readJsonDataStore()` references which are plenty and can be found in multiple mining modules?

Patents and covid19 related tests used many resource files loaded using classloader. That's why there are a lot of changes in related classes.  I've run parallel jenkins builds and they succeeded, so all the broken parts are fixed.

iis-common/src/main/java/eu/dnetlib/iis/common/utils/ResourceUtils.java
```diff
@@ -0,0 +1,25 @@
...
+public class ResourceUtils {
+
+    /**
+     * Returns a resource path with properly handled special characters.
+     *
+     * @param cl   ClassLoader instance for resource extraction.
+     * @param name Path of the resource
+     * @return String representing the path to the resource.
+     */
+    public static String resourcePath(ClassLoader cl, String name) {
```
marekhorst, line 18: A little bit of nitpicking: I would stick to a methods naming convention and specify the action that is being undertaken (missing an appropriate verb).

marekhorst (changes requested):
>> What is so special about the patents and covid19 algos those are the ones mostly affected? Speaking other words: does it mean we don't need to fix other `JsonAvroTestUtils.readJsonDataStore()` references which are plenty and can be found in multiple mining modules?

> Patents and covid19 related tests used many resource files loaded using classloader. That's why there are a lot of changes in related classes. I've run parallel jenkins builds and they succeeded, so all the broken parts are fixed.

OK, so we have to deal only with the places involving reading resources using `ClassLoader` while in all the other `JsonAvroTestUtils.readJsonDataStore()` references where we provide relative path we don't need to fix anything, right?

Does it mean if we align the code involving `ClassLoader` with the rest by simply adding `src/test/resources/` to the path (and removing `ClassLoader` involvement) then we would automatically get rid of this problem without normalizing resource path? I don't ming alignment ;)

mpol (approved):
LGTM

przemyslawjacewicz:
> > > What is so special about the patents and covid19 algos those are the ones mostly affected? Speaking other words: does it mean we don't need to fix other `JsonAvroTestUtils.readJsonDataStore()` references which are plenty and can be found in multiple mining modules?
> 
> > Patents and covid19 related tests used many resource files loaded using classloader. That's why there are a lot of changes in related classes. I've run parallel jenkins builds and they succeeded, so all the broken parts are fixed.
> 
> OK, so we have to deal only with the places involving reading resources using `ClassLoader` while in all the other `JsonAvroTestUtils.readJsonDataStore()` references where we provide relative path we don't need to fix anything, right?
> 
> Does it mean if we align the code involving `ClassLoader` with the rest by simply adding `src/test/resources/` to the path (and removing `ClassLoader` involvement) then we would automatically get rid of this problem without normalizing resource path? I don't mind alignment ;)

The fix to failing jenkins build seemed quick and easy at first but it is turning into a more generic case now ;-) The core problem of failing builds due to special characters in resource path is fixed with this PR but I'm not sure if this is the solution we should accept. Looking more closely into the problem I've noticed that there is a bit of a mess with getting resources from classpath. We have 12 distinct 'ways' of getting resource content! This is definitely a place for unification and alignment by providing a single utility class for extraction of resources that could be applied to all related code. However, this will exceed the context of the source issue #1134. It would be much better to create a separate issue for unification of resource handling and discuss the solution there. 

As for this PR we could 
* accept it as it is now and provide a generic solution for resource handling with next PR
* work on the generic solution within this PR
* close this PR without merging and take the outcome of this PR to providing the generic solution within next PR

I would go with the third option to avoid messing up the `master` with code that would be changed or removed in a bit. @mpol  and @marekhorst what do you think ?

marekhorst:
> I would go with the third option to avoid messing up the `master` with code that would be changed or removed in a bit. @mpol and @marekhorst what do you think ?

Agree.

mpol:
I think that either 1 or 3 makes sense. It depends on how long will the new PR take and how many build failures can we expect in the meantime.

przemyslawjacewicz:
I'm closing this PR without merging and taking code review comments as input for a more generic solution proposed in PR #1152 .

PR #1152, opened 2020-09-18
Closes #1134: Introduce a common functionality for retrieving resources

### LINKED ISSUE #1134: Fix multibranch build failing when more than one build is being run in parallel

przemyslawjacewicz:
Inside this PR
* introduction of `eu.dnetlib.iis.common.StaticResourceProvider` with utility methods for handling resources on classpath: finding resource path and resource content
* making use of `eu.dnetlib.iis.common.StaticResourceProvider` for handling resources
* various cleanups due to new utility class
  * removal of unnecessary methods and code parts for retrieving resource paths or contents in production code and in tests
  * fixes to related code parts e.g. description updates

Note: `eu.dnetlib.iis.common.StaticResourceProviderTest` uses a new test method naming convention: `public void given_when_then()`. I've included it as an experiment, I think it's quite convenient.  However, if opinions will differ I can change it to our usual style.


iis-common/src/main/java/eu/dnetlib/iis/common/StaticResourceProvider.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class StaticResourceProvider {
```
marekhorst, line 14: I guess we could add a brief javadoc description for this class.

And one minor remark: since all the methods from this class are relying on classpath resources then maybe it would more informative when renaming `StaticResourceProvider` to `ClassPathResourceProvider`. What do you think?

marekhorst (changes requested):
Just a few minor remarks on potential improvements.

iis-common/src/main/java/eu/dnetlib/iis/common/StaticResourceProvider.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+    }
+
+    /**
+     * Returns the input stream reader of a resource on the classpath.
+     *
+     * @param location String indicating the location of resource on the classpath.
+     * @return InputStreamReader of the resource.
+     */
+    public static InputStreamReader getResourceInputStreamReader(String location) {
```
mpol, line 76: Maybe `Reader getResourceReader(String location)`?

iis-common/src/main/java/eu/dnetlib/iis/common/StaticResourceProvider.java
```diff
@@ -0,0 +1,79 @@
...
+    public static InputStream getResourceInputStream(String location) {
+        try {
+            return new ClassPathResource(location).getInputStream();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Returns the input stream reader of a resource on the classpath.
```
mpol, line 71: Maybe it should mention that it uses UTF-8?

iis-common/src/main/java/eu/dnetlib/iis/common/StaticResourceProvider.java
mpol: +1 for Javadoc
The `StaticResourceProvider` name at least would need some explanation. And I have no intuition if `ClassPathResourceProvider` is right or too much of an implementation-detail name but all the docs below mention classpath, so maybe it would be a better name?

iis-wf/iis-wf-affmatching/src/test/java/eu/dnetlib/iis/wf/affmatching/AffMatchingJobTest.java
```diff
@@ -82,15 +83,22 @@ public void affiliationMatchingJob() throws IOException {
...
+        String jsonInputInferredDocProjPath = StaticResourceProvider
+                .getResourcePath("data/input/docProjInferred.json");
+        String jsonInputDocProjPath = StaticResourceProvider
+                .getResourcePath("data/input/docProj.json");
+        String jsonInputProjOrgPath = StaticResourceProvider
+                .getResourcePath("data/input/projOrg.json");
+
+        String jsonOutputPath = StaticResourceProvider
+                .getResourcePath("data/expectedOutput/matchedOrganizations.json");
+        String jsonOutputReportPath = StaticResourceProvider
```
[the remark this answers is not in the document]
przemyslawjacewicz: Yes, you're right, but I'd say that it's not a real drawback and can add to code readability. Bounding resources to packages makes clear where the resource is used.

iis-common/src/main/java/eu/dnetlib/iis/common/StaticResourceProvider.java
przemyslawjacewicz: Yes, I thought of renaming the class to :) Plus adding class javadoc.

przemyslawjacewicz:
I've pushed code review changes.

marekhorst (approved):
LGTM.

PR #1153, opened 2020-09-22
Improve the "text-snippets".

LSmyrnaios:
Make more clear what the match is and increase performance by using the SQLite-string-concatenation.


LSmyrnaios:
Replaced by https://github.com/openaire/iis/pull/1165

PR #1156, opened 2020-09-23
Closes #1102: Drop redundant delete operations from workflow definitions of spark actions

### LINKED ISSUE #1102: Drop redundant delete operations from workflow definitions of spark actions
There are `<delete>` elements defined within `<prepare>` section of spark action definitions within `workflow.xml` files of various mining modules which are redundant and are overlapping with similar operations defined in spark code of underlying spark jobs.

Historically speaking the delete operations within `workflow.xml` definitions were defined first and are defined also for different kinds of oozie actions. The problem was the spark retry mechanism: once it was triggered it did not execute the actions from the `<prepare>` section within oozie wf definition because it was handled on a different operational (spark) level. So the purging of output/report directories had to be encoded as the very first operations within spark code.

przemyslawjacewicz:
This PR removes `prepare` statements for spark actions. It also adds minor fixes for related code: missing `HdfsUtils.remove` and using factory methods for creating spark context.

### DIFF
--- a/iis-wf/iis-wf-affmatching/src/main/resources/eu/dnetlib/iis/wf/affmatching/dedup/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-affmatching/src/main/resources/eu/dnetlib/iis/wf/affmatching/dedup/oozie_app/workflow.xml
@@ -84,12 +84,6 @@
 
     <action name="affmatching_dedup">
         <spark xmlns="uri:oozie:spark-action:0.2">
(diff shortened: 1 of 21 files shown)


marekhorst (approved):
LGTM.

przemyslawjacewicz:
For reference: no factory methods were used for creating spark context. The description refers to initial version of this PR with breaking changes in citation matching module. These changes were reverted in the final version.

PR #1158, opened 2020-09-24
Closes #1151: Introduce a common functionality for retrieving resources - FIX

### LINKED ISSUE #1151: Introduce a common functionality for retrieving resources
This issue is a followup of issue #1134 .

While working on #1134 I've found that we have quite a few different 'ways' of getting resources for tests, with one way that does not work when there are special characters in resource path e.g. `@`. Introducing a common handling of resources would therefore be very helpful to avoid this kind of problems in the future.

The goal of this issue is to create this common functionality and firstly apply it to failing tests. The functionality can then be safely used for future tests. Refactorization of all tests with resources is not necessary at this point.

For reference these are the 'ways' of getting resources used in IIS
* `eu.dnetlib.iis.wf.importer.StaticResourcesProvider` : 
```java
    public static List<String> getResources(String ...resourcesLocations) {
        List<String> resources = Lists.newArrayList();
        for (String resourceLocation : resourcesLocations) {
            try (InputStream input = StaticResourcesProvider.class.getResourceAsStream(resourceLocation)) {
                resources.add(IOUtils.toString(input, "utf8"));
            } catch (IOException e) {
                throw new RuntimeException("Unable to read resource: " + resourceLocation, e);
            }
        }
        return resources;
    }
```
* `Thread.currentThread().getContextClassLoader().getResourceAsStream(path)`
* using Spring `org.springframework.core.io.ResourceLoader`
 ```java
 ResourceLoader loader = new DefaultResourceLoader();
 Resource resource = loader.getResource(location);
 ```
* `Objects.requireNonNull(classLoader.getResource(name)).toURI().getPath()`
* `getClass().getResourceAsStream(classname)`
* `getClass().getResource(resource).toURI()`
* `TestingConsumer.class.getResourceAsStream()`
 * `DocumentClassificationJobTest.class.getResource(path).getPath()`
 * using spring `org.springframework.core.io.support.ResourcePatternResolver`
 ```java
 ResourcePatternResolver resolver = new PathMatchingResourcePatternResolver();
 Resource[] resources = resolver.getResources(ResourcePatternResolver.CLASSPATH_ALL_URL_PREFIX + directoryToScan + "**");
 ```
 * using this snippet
```java
private ClassLoader cl = getClass().getClassLoader();
Objects.requireNonNull(cl.getResource(path)).getFile();
```
 * `ClassLoader.class.getResourceAsStream(path)`
 * `getClass().getClassLoader().getResource(path)).getFile()`
 * providing the path explicitely with `src/test/resources` prefix

przemyslawjacewicz:
This PR fixes a bug in `eu.dnetlib.iis.common.java.jsonworkflownodes.ClassPathResourceToHdfsCopier` introduced with merge of PR #1152 when `parameters.get(PARAM_INPUT_CLASSPATH_RESOURCE)` was removed when creating resource input stream. Test was also added for the referenced class.

### DIFF
--- a/iis-common/src/main/java/eu/dnetlib/iis/common/java/jsonworkflownodes/ClassPathResourceToHdfsCopier.java
+++ b/iis-common/src/main/java/eu/dnetlib/iis/common/java/jsonworkflownodes/ClassPathResourceToHdfsCopier.java
@@ -4,6 +4,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Function;
 
 import eu.dnetlib.iis.common.ClassPathResourceProvider;
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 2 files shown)


marekhorst (approved):
LGTM.

PR #1160, opened 2020-09-28
Closes #1151: Introduce a common functionality for retrieving resources - FIX#2

### LINKED ISSUE #1151: Introduce a common functionality for retrieving resources

przemyslawjacewicz:
This PR fixes a failing integration test suite `eu.dnetlib.iis.wf.referenceextraction.softwareurl.SoftwareUrlRefExtractionWfTest`. It seems that as a result of merging #1152 integration tests timeout value needs to be increased. I've increased it to 30min/1800s.

Also a minor fix - removal of `throws` clause because no exceptions are thrown in test methods.

### DIFF
--- a/iis-wf/iis-wf-referenceextraction/src/test/java/eu/dnetlib/iis/wf/referenceextraction/softwareurl/SoftwareUrlRefExtractionWfTest.java
+++ b/iis-wf/iis-wf-referenceextraction/src/test/java/eu/dnetlib/iis/wf/referenceextraction/softwareurl/SoftwareUrlRefExtractionWfTest.java
@@ -1,5 +1,6 @@
 package eu.dnetlib.iis.wf.referenceextraction.softwareurl;
 
+import eu.dnetlib.iis.common.OozieWorkflowTestConfiguration;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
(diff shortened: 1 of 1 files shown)


marekhorst (approved):
LGTM. Although it looks pretty strange software mining integration test takes more than 10 minutes. Probably handling the cache introduces quite heavy overhead...

OK, I just verified that, it seems even the simple postprocessing phase takes 10 minutes:

http://iis-ci-test.ocean.icm.edu.pl:19888/jobhistory/job/job_1601296763140_0027

mainly because there are static 1000 tasks which is not suitable for an integration test.

If we are looking for optimizing test execution time we could override those kind of parameters to smaller numbers. I will create dedicated ticket for this.

Issue #1066, opened 2020-04-20
Remove unused dnet dependencies


marekhorst:
While working on #1045 and #1051 I have noticed several dependencies which seem to be unused and could be removed. 

Those two:
* `dnet-openaire-data-protos`
* `dnet-actionmanager-common`

including an old Oaf model and classes related to an old actionmanager service were removed as a part of #1065 pull request (introducing exporter module refactoring: #1045) but there are still some libs which could potentially be removed:
* `cnr-rmi-api` (according to `pom.xml` comments: MDStore related which is not used anymore)
* `cnr-service-utils` (according to `pom.xml` comments: MDStore related which is not used anymore)
* `cnr-resultset-client` (ResultSet related, need to check whether we still need this)
* `xalan` (seem to be related to an old export procedure requiring XML transformation)

which are not related to Oaf model refactoring so it should be considered as a separate task.

marekhorst:
There is one another dnet dependency, `dnet-openaireplus-mapping-utils`, which is used only in a limited scope so we could consider removing it after getting rid of the following references:
* `eu.dnetlib.data.transform.xml.AbstractDNetXsltFunctions#md5()`
* `eu.dnetlib.data.transform.xml.AbstractDNetXsltFunctions#oafId()`
* `eu.dnetlib.openaire.exporter.model.ProjectDetail`

If it turns out to require too much effort we could leave it as it is.

przemyslawjacewicz:
The list of removed dependencies
* `dnet-openaireplus-mapping-utils`
* `cnr-rmi-api`
* `cnr-service-utils`
* `xalan`

`cnr-resultset-client` is used and its usage is not easily replacable, so I'm leaving it.

Issue #1079, opened 2020-05-15
Remove dnet-openaireplus-mapping-utils dependency


marekhorst:
After abandoning an old protocol buffer model and shifting towards new oaf dhp model (#1045 and #1051) we could get rid of some of the dnet dependencies revolving around the old model. 

This was partially done by removing `dnet-openaire-data-protos` explicitly defined as direct dependency which came back as transitive one because of `dnet-openaireplus-mapping-utils` dependency which seems also to be mostly related to the old protobuf model.

We should get rid of `dnet-openaireplus-mapping-utils` dependency which will result in dropping several other unwanted dependencies such as `dnet-openaire-data-protos` and `dnet-pace-core` including pretty long external dependency tail.

After inspecting IIS source code I have found two groups of references to classes from `dnet-openaireplus-mapping-utils`:
* `eu.dnetlib.data.transform.xmlAbstractDNetXsltFunctions.md5()` in several places within exporter module
* `eu.dnetlib.openaire.exporter.model.ProjectDetail` in projects importer module relying on RESTful streaming endpoint

The first one should be quite easily replaceable while the second one might require additional effort because it is used to deserialize JSON records obtained from projects streaming endpoint.

marekhorst:
> After inspecting IIS source code I have found two groups of references to classes from `dnet-openaireplus-mapping-utils`:
> 
> * `eu.dnetlib.data.transform.xmlAbstractDNetXsltFunctions.md5()` in several places within exporter module
> * `eu.dnetlib.openaire.exporter.model.ProjectDetail` in projects importer module relying on RESTful streaming endpoint
> 
> The first one should be quite easily replaceable while the second one might require additional effort because it is used to deserialize JSON records obtained from projects streaming endpoint.

@claudioatzori do you also see the `dnet-openaireplus-mapping-utils` as something we could get rid of while abandoning an old protobuf Oaf model? If this is the case do you plan to move e.g. projects streaming endpoint model (mentioned above) to some other module? I know you have plenty to do so if you come to any kind of refactoring in near future just keep my note in your mind :)

dnet-team:
The ProjectDetail bean can probably be moved in the `dhp-schema` module; It is already the IIS reference module for the recent D-Net common model definitions, so it could be the target of a simple refactoring. I only need to check which other model classes could be moved there. I will give you a note as soon as I manage to progress on this front.

claudioatzori:
[...]

PR #1162, opened 2020-09-29
Closing #1066 and #1079: DNet dependency cleanup

przemyslawjacewicz:
This PR closes issues #1066 and #1079 . Changes overview
* removal of dependencies 
* `AbstractDNetXsltFunctions.md5()` replaced by Apache Commons `DigestUtils.md5Hex()`
* `AbstractDNetXsltFunctions.oafId()` moved to ISS with minor improvements
* `ProjectDetail` moved to IIS

### DIFF
--- a/iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/stream/project/ProjectDetailConverter.java
+++ b/iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/stream/project/ProjectDetailConverter.java
@@ -2,15 +2,14 @@ package eu.dnetlib.iis.wf.importer.stream.project;
 
 import java.io.IOException;
 
+import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 
-import eu.dnetlib.data.transform.xml.AbstractDNetXsltFunctions;
 import eu.dnetlib.iis.common.InfoSpaceConstants;
 import eu.dnetlib.iis.importer.schemas.Project;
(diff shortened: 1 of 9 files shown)


iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/stream/project/ProjectDetailConverter.java
```diff
@@ -63,8 +62,17 @@ private static String normalizeId(String sourceId) {
...
+
+    private static String projectOafId(String prefix, String id) {
+        if (id.isEmpty() || prefix.isEmpty()) {
+            return "";
+        }
+        return projectOafSimpleId(prefix + "::" + DigestUtils.md5Hex(id));
+    }
+
+    private static String projectOafSimpleId(String id) {
+        return (InfoSpaceConstants.ROW_PREFIX_PROJECT + id).replaceAll("\\s|\\n", "");
```
marekhorst, line 76: Again, this is kind of similar to `ProjectDetail` case: I am afraid we could start producing invalid identifiers once the logic gets changed on DNet side. This could be pretty difficult to spot and debug.

I would prefer to "borrow" this logic for a short period of time and rely on a common codebase once it gets published e.g. in some new lightweight dependency (like `dhp-schema` for `ProjectDetail`). So we should probably ask Claudio if it would be possible to expose this method in a different package than `dnet-openaireplus-mapping-utils` which is pretty heavy.

marekhorst (approved):
Approving, but we should make sure we can switch back to DNet codebase once it gets released. So please create a dedicated ticket so we won't lose it from our radar.

przemyslawjacewicz:
I've created an issue for reintroducing DNet functionalities related to `ProjectDetail`: #1163 .

PR #1155, opened 2020-09-23
Closes #1071 and #1103: Add repartition with number of files for datastore write 

przemyslawjacewicz:
This PR closes #1071 and #1103. 

Changes overview:
- replacing `coalesce()` with `repartition()` when writing datastores with desired number of emitted files; for writing report datastores repartitioning to a single partition is added
- adding relevant test assertions for produced number of files
- adding a general HDFS file counting method: `HdfsUtils.countFiles`
- minor fixes and changes for related files
  -  changing from `com.google.common.io.Files` to `java.nio.file.Files` - this needs a bit of clarification because it is not directly related to the goal of this PR: at first I've used `java.nio.file.Files.list` for listing and counting files in tests, this did not go well with `com.google.common.io.Files` so I've removed the google version, then I've added `HdfsUtils.countFiles` and changes related to `java.nio.file.Files` could be reverted but I've decided to keep the changes because `com.google.common.io.Files` gives warnings when `createTempDir()` is used `'createTempDir()' is declared in unstable class 'com.google.common.io.Files' marked with @Beta`
  - for test with number of emitted files I've changed the number of emitted files from `1` to `2` to make it different then the number for reports

Issue #1103, opened 2020-07-10
Investigate the impact of setting the number of emitted files on jobs performance


przemyslawjacewicz:
Many IIS workflows limit the number of files created when saving datastores to HDFS. The initial motivation for this behavior was to not allow the file number to grow beyond reasonable size when unioning with other datastores. In spark jobs the limit of created files is achieved by using `coalesce` with desired number of files as argument. This however has an effect on the number of tasks spawned in the stage performing the actual write - the number of tasks is limited by the number of files to be created. This behavior can hit the performance if the number of tasks is low and does not utilize the cluster properly.

The main purpose of this task is to find the workflows limiting the number of emitted files and investigate the impact of the number of files on job performance.

marekhorst:
> This however has an effect on the number of tasks spawned in the stage performing the actual write - the number of tasks is limited by the number of files to be created.

This gave me this recollection regarding the TARA mining: probably the number of the tasks executing the madis is equal to the number of partitions provided as `coalesce()` parameter because the execution plan bound madis execution and write operations together in a single stage. If it was at different stage the number of madis tasks would probably differ from the number of output files.

przemyslawjacewicz:
Yes, that's right. Pipe description says
```
With pipe, you can return an RDD created by piping elements to a forked external process. The resulting RDD is computed by executing the given process once per partition. All elements of each input partition are written to a process’s stdin as lines of input separated by a newline. The resulting partition consists of the process’s stdout output, with each line of stdout resulting in one element of the output partition. A process is invoked even for empty partitions.
```
So there is a one-to-one correspondence between number of tasks and number of partitions running madis. Because coalesce was pipelined into the same stage as madis execution the number of output files ultimately sets the number of tasks running madis.

marekhorst:
OK, I think I have finally found the solution. In `coalesce()` method documentation:

https://spark.apache.org/docs/2.4.0/api/python/pyspark.sql.html#pyspark.sql.DataFrame.coalesce

and:

https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L494

one can read:

> However, if you’re doing a drastic coalesce, e.g. to numPartitions = 1, this may result in your computation taking place on fewer nodes than you like (e.g. one node in the case of numPartitions = 1). To avoid this, you can call repartition(). This will add a shuffle step, but means the current upstream partitions will be executed in parallel (per whatever the current partitioning is).

Some interesting remarks are also available in those posts:
[...]

PR #1168, opened 2020-10-05
Closes #1071 and #1103: Add repartition with number of files for datastore write - FIX#1 

### LINKED ISSUE #1071: Decrease number of output files produced for reports
This task is similar to #923.

The number of files generated for reports can be unnecessary large with only single file containing the results. Due to the small amount of generated report entries we can decrease the number of files to a single file.

The reports with large number of files, indicated as subdirs in main `report` dir
* import_patent
* aff-matching
* affmatching-dedup
* citationmatching_direct
* citationmatching_fuzzy
* document_classification
* document_to_community_by_root_conceptid
* document_to_concept_by_root_conceptid
* document_to_project_by_funder
* project-based-docorg-matching
* referenceextraction_software
* export_software

przemyslawjacewicz:
This PR fixes failing integration tests bug introduces after merging #1155 . Introducing explicit repartition step when writing datastores causes datastores to be always saved with explicit number of files even if initial number of partitions is smaller. In tests we finish with a large number of empty files that are subsequently processed using for example madis processes. So instead of a few map tasks to process the files we have a large number of unnecessary tasks. Having to process this large number of tasks vastly increases the overall execution time of an integration test. The solution in this PR is to decrease the number of output files to 1 for tests. This however causes a few more changes in related workflows to properly name the parameters.

### DIFF
--- a/iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
@@ -482,6 +482,22 @@
                     <name>patentServiceEndpointOpsUriRoot</name>
                     <value>expectedOpsUriRoot</value>
                 </property>
+				<property>
+					<name>affiliation_matching_number_of_emitted_files</name>
+					<value>1</value>
+				</property>
+				<property>
+					<name>documentsclassification_main_number_of_partitions</name>
+					<value>1</value>
+				</property>
+				<property>
+					<name>referenceextraction_patent_main_number_of_emitted_files</name>
+					<value>1</value>
+				</property>
+				<property>
+					<name>webcrawlNumberOfEmittedFiles</name>
+					<value>1</value>
+				</property>
 			</configuration>
 		</sub-workflow>
 		<ok to="consumer" />
--- a/iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
@@ -487,10 +487,6 @@
 					<value>1</value>
 				</property>
 				<property>
-					<name>documentsclassification_main_number_of_partitions</name>
-					<value>1</value>
-				</property>
-				<property>
 					<name>referenceextraction_patent_number_of_emitted_filesa</name>
 					<value>1</value>
 				</property>
--- a/iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
@@ -491,7 +491,7 @@
 					<value>1</value>
 				</property>
 				<property>
-					<name>referenceextraction_patent_main_number_of_emitted_files</name>
+					<name>referenceextraction_patent_number_of_emitted_filesa</name>
 					<value>1</value>
 				</property>
 				<property>
--- a/iis-wf/iis-wf-affmatching/src/main/resources/eu/dnetlib/iis/wf/affmatching/main/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-affmatching/src/main/resources/eu/dnetlib/iis/wf/affmatching/main/oozie_app/workflow.xml
@@ -92,7 +92,7 @@
             <description>spark 2.* event log dir location</description>
         </property>
         <property>
-            <name>number_of_emitted_files</name>
+            <name>affiliation_matching_number_of_emitted_files</name>
             <value>1000</value>
             <description>number of files created by webcrawler module</description>
         </property>
@@ -151,7 +151,7 @@
             <arg>-outputAvroReportPath=${output_report_root_path}/aff-matching</arg>
             
             <arg>-inputDocProjConfidenceThreshold=${document_project_confidence_threshold}</arg>
-            <arg>-numberOfEmittedFiles=${number_of_emitted_files}</arg>
+            <arg>-numberOfEmittedFiles=${affiliation_matching_number_of_emitted_files}</arg>
             
         </spark>
         <ok to="end" />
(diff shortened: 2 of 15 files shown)


iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
```diff
@@ -482,6 +482,22 @@
                     <name>patentServiceEndpointOpsUriRoot</name>
                     <value>expectedOpsUriRoot</value>
                 </property>
+				<property>
+					<name>affiliation_matching_number_of_emitted_files</name>
+					<value>1</value>
+				</property>
+				<property>
+					<name>documentsclassification_main_number_of_partitions</name>
```
marekhorst, line 490: Could we align this property name with all the other cases by renaming:

`documentsclassification_main_number_of_partitions`

into:

`documentsclassification_number_of_emitted_files`

so I would skip the unneccessary `_main` part and rename `partitions` to `emitted_files`.

iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
```diff
@@ -482,6 +482,22 @@
...
+					<value>1</value>
+				</property>
+				<property>
+					<name>referenceextraction_patent_main_number_of_emitted_files</name>
+					<value>1</value>
+				</property>
+				<property>
+					<name>webcrawlNumberOfEmittedFiles</name>
```
marekhorst, line 498: This obviously caught my eye (camel vs snake case) but I guess changing this convention here would force us to change the convention for all the webrawl params and it will become slightly off the scope of this task. So I guess we should leave it as it is and maybe commit some time to align on the convention later on.

iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
```diff
@@ -482,6 +482,22 @@
...
+				<property>
+					<name>affiliation_matching_number_of_emitted_files</name>
+					<value>1</value>
+				</property>
+				<property>
+					<name>documentsclassification_main_number_of_partitions</name>
+					<value>1</value>
+				</property>
+				<property>
+					<name>referenceextraction_patent_main_number_of_emitted_files</name>
```
marekhorst, line 494: Could we make it shorter, e.g. like `referenceextraction_patent_number_of_emitted_files`?

marekhorst (changes requested):
Just a minor remarks on parameters naming.

iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
przemyslawjacewicz: I've just realized that the original change may be a bit far fetched. The parameter does not directly correspond to a repartition at data store write. The repartition happens a few transformations earlier and is followed by maps, filters and a pipe. All those transformations preserve the number of partitions so in the end the parameter is also responsible for the number of emitted files, but this is not that obvious at the point where this parameter is used. Also adding any wide transformation in between will result in different number of emitted files. So I think we can
* remove this parameter altogether - it seems that this parameter was not used before, the default value is `$UNDEFINED$` which corresponds to skipping repartition, as the result non-test execution would remain the same and test execution would be done with small amount of partitions  corresonding to test data
* keep the original parameter and introduce another parameter for the number of emitted files and set both parameters to 1 for tests
* move the repartition to the point where the datastore write happens and change the name of the parameter - this will make the name appropriate and will also set the number of partitions for pipe just as now 
* change the name of the parameter as you propose and set the parameter value to 1 for tests
* keep the implementation as it is now without name change with setting the parameter value to 1 for tests

I'd go with the first option because it removes unused parameter and does not introduce any new ones. What do you think ?

iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
przemyslawjacewicz: This seems like the best option for now - changing all the names would be quite messy now.

przemyslawjacewicz:
The full set of integration tests passed.

iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
marekhorst: It's good you have reminded about those transformations: remembering this I am rolling back the potential improvement of renaming this param into `documentsclassification_number_of_emitted_files`. 

The reason this parameter was introduced was to cope with the long lasting attempts. We had to increase the number of partitions in order to increase the total number of tasks doing the job. This way the scheduler can juggle the cluster resources more efficiently.

marekhorst (changes requested):
LGTM although as described in the last comment - we could get rid of explicitly defined `documentsclassification_main_number_of_partitions` parameter in all the test workflows.

iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/processing/sampledataproducer/oozie_app/workflow.xml
przemyslawjacewicz: My original idea was to completely remove the parameter but as you point out we definitely need it ;) I'm removing the parameter only from tests then.

przemyslawjacewicz:
> LGTM although as described in the last comment - we could get rid of explicitly defined `documentsclassification_main_number_of_partitions` parameter in all the test workflows.

I've removed the parameter from test workflows, re-requesting the review.

przemyslawjacewicz:
> LGTM although as described in the last comment - we could get rid of explicitly defined `documentsclassification_main_number_of_partitions` parameter in all the test workflows.

Removing `documentsclassification_main_number_of_partitions` parameter from tests was part of 'LGTM' message so I think an explicit review can be skipped and this PR can be merged.

PR #1172, opened 2020-10-18
Closes #1159: Citation matching stage occupies resources for a very long time 

### LINKED ISSUE #1159: Citation matching stage occupies resources for a very long time
I just noticed the citation matching algorithm occupies most of the cluster resources (capped at 80% of total resources by scheduler but still) during one of its stages:

http://iis-cdh5-test-m1.ocean.icm.edu.pl:8088/proxy/application_1595483222400_14384/stages/stage?id=21&attempt=0

namely `mapPartitionsToPair at BestMatchedCitationPicker.java:42` which lasts for over 4h (and counting). This makes it very difficult to recollect resources for other jobs including analytics.

We should take a closer look at the calculations being made and optimize it. One of the approaches could be increasing the number of partitions (currently set to 1000) for this stage to get more tasks running this job (thus decreasing the amount of time each stage lasts).

We should also take a closer look at other mining modules and search for similar cases. Most of them were already eliminated by #991.

przemyslawjacewicz:
This PR removes hardcoded value for citation matching number of partitions and moves the parameter to IIS deployment  repository. As a result the value of this parameter will be set upon deployment of IIS. Therefore this PR is matched by PR in IIS deployment repo.

### DIFF
--- a/iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/main/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/main/oozie_app/workflow.xml
@@ -203,6 +203,11 @@
                 Set to 'a^' by default to guarantee nothing will be matched.
             </description>
         </property>
+        <!-- citation matching related -->
+        <property>
+            <name>citationmatchingFuzzyNumberOfPartitions</name>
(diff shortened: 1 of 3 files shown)


przemyslawjacewicz:
IIS deployment repo PR: [5](https://git.icm.edu.pl/openaire/iis-deployment/-/merge_requests/5).

marekhorst (approved):
LGTM.

PR #1177, opened 2020-10-23
Closes #1176: Change the order of distcp and pushing metrics operations during the export phase

### LINKED ISSUE #1176: Change the order of distcp and pushing metrics operations during the export phase
It looks like the `distcp` operation responsible for copying the mining outcome to the Action Manager HDFS directory fails quite often recently, mainly due to denied permission to write in an output directory. 

This is a result of misconfiguration (either lack of previledges or missing directory) and each time should be solved anyway but the problem is it causes the metrics not to be exported after the IIS primary_export phase which requires manual export.

We could reorder this two operations within the oozie workflow and let exporting the metrics first and the do the `distcp` operation so even when it fails we still get the metrics properly exported.

marekhorst:
Simply swapping the order.


przemyslawjacewicz (approved):
LGTM

Issue #1173, opened 2020-10-19
java.net.ConnectException in integration tests


przemyslawjacewicz:
Integration tests log is filled with these messages
```text
20/10/17 05:06:06 WARN ClientCnxn: Session 0x10d72405cd90000 for server localhost/127.0.0.1:44927, unexpected error, closing socket connection and attempting reconnect
java.net.ConnectException: Connection refused
	at sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
	at sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:717)
	at org.apache.zookeeper.ClientCnxnSocketNIO.doTransport(ClientCnxnSocketNIO.java:355)
	at org.apache.zookeeper.ClientCnxn$SendThread.run(ClientCnxn.java:1236)
```
Test runs are successful but logs are unreadable. We should find the reason for these exceptions and fix it.

przemyslawjacewicz:
The first occurrance of this error is after execution of `eu.dnetlib.iis.wf.referenceextraction.softwareurl.CachedWebCrawlerJobTest` which uses `org.apache.curator.test.TestingServer` which in turn internally uses `TestingZooKeeperServer`. So the error seems to be connected with `org.apache.curator.test.TestingServer`. Trying to replicate the error by executing this single test in Jenkins fails but execution in local environment shows this error. Here's the log from local execution
```text
20/11/06 13:47:51 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
20/11/06 13:48:44 INFO ZookeeperLockManagerFactory: initializing root node: /cache
20/11/06 13:48:44 INFO ZookeeperLockManagerFactory: root node initialized
20/11/06 13:48:45 INFO LockManagingProcessUtils: trying to obtain lock: /cache/_tmp_junit6375610101001173385_spark_webcrawler_cache
20/11/06 13:48:45 INFO LockManagingProcessUtils: lock not found, creating new lock
20/11/06 13:48:45 INFO LockManagingProcessUtils: creating new lock instance: /cache/_tmp_junit6375610101001173385_spark_webcrawler_cache
20/11/06 13:48:45 INFO LockManagingProcessUtils: lock/cache/_tmp_junit6375610101001173385_spark_webcrawler_cache created
Warning: fs.defaultFS is not set when running "chmod" command.
[...]

PR #1187, opened 2020-11-06
#1173: java.net.ConnectException in integration tests

przemyslawjacewicz:
This PR is an attempt to eliminate `java.net.ConnectException` from integration tests. Inside this PR we're converting instances of `org.apache.curator.test.TestingServer` used in tests to static variables and using `close()` rather than `stop()` for server termination. Local testing shows that this may be the cause of the exceptions. However, further changes may be required and integration tests log must be checked for exceptions afterwards.

Linked issue: #1173 .


marekhorst (approved):
LGTM, although I assume `zookeperServer#stop()` -> `zookeeperServer#close()` change alone did not fix the issue, so it was introduced to free resources in a better way, right?

przemyslawjacewicz:
> LGTM, although I assume `zookeperServer#stop()` -> `zookeeperServer#close()` change alone did not fix the issue, so it was introduced to free resources in a better way, right?

Yes, you're right. Actually in my local tests making `TestingServer` static was enough to fix the exception but I've also changed `stop()` to `close()` as a precaution because close doc says
```text
Close the server and any open clients and delete the temp directory
```
So I though it is also a good idea to use `close()`. I'm still not 100% sure these changes will fix the problem, so integration tests log needs  to be under obervation ;-)

PR #1189, opened 2020-11-12
Closes #1184: Refactor tests with SparkSession to use common base class for SparkSession management

### LINKED ISSUE #1184: Refactor tests with SparkSession to use common base class for SparkSession management
The normal approach for creating `SparkSession` in tests is to create it manually in 'before' methods and close it in 'after' methods. This approach can be converted to more generic one by creating a test base class that takes care of creating and disposing a `SparkSession` instance. `TestWithSharedSparkSession` introduced in PR #1179 exactly does this for `SparkSession`. The goal of this issue is to apply this approach to all tests creating `SparkSession`.

przemyslawjacewicz:
This PR closes #1184 .

Each test extending `eu.dnetlib.iis.common.spark.TestWithSharedSparkSession` will automatically be tagged as slow test.

One thing to note is that if a test extending `eu.dnetlib.iis.common.spark.TestWithSharedSparkSession` has also `BeforeEach` method then test's `BeforeEach` method must call base class `BeforeEach` method by using `super.beforeEach()`.

### DIFF
--- a/iis-common/src/test/java/eu/dnetlib/iis/common/spark/TestWithSharedSparkSession.java
+++ b/iis-common/src/test/java/eu/dnetlib/iis/common/spark/TestWithSharedSparkSession.java
@@ -13,7 +13,7 @@ import java.util.Objects;
  */
 @SlowTest
 public class TestWithSharedSparkSession {
-    private static transient SparkSession _spark;
+    private static SparkSession _spark;
     protected boolean initialized = false;
 
     public SparkSession spark() {
@@ -26,6 +26,7 @@ public class TestWithSharedSparkSession {
 
         if (!initialized) {
             SparkConf conf = new SparkConf();
+            conf.setAppName(getClass().getSimpleName());
             conf.setMaster("local");
             conf.set("spark.driver.host", "localhost");
             conf.set("spark.serializer", "org.apache.spark.serializer.KryoSerializer");
--- a/iis-common/src/test/java/eu/dnetlib/iis/common/spark/TestWithSharedSparkSession.java
+++ b/iis-common/src/test/java/eu/dnetlib/iis/common/spark/TestWithSharedSparkSession.java
@@ -1,5 +1,6 @@
 package eu.dnetlib.iis.common.spark;
 
+import eu.dnetlib.iis.common.SlowTest;
 import org.apache.spark.SparkConf;
 import org.apache.spark.sql.SparkSession;
 import org.junit.jupiter.api.AfterAll;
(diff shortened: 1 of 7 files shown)


iis-common/src/test/java/eu/dnetlib/iis/common/spark/TestWithSharedSparkSession.java
```diff
@@ -26,6 +28,7 @@ public void beforeEach() {
             SparkConf conf = new SparkConf();
             conf.setMaster("local");
             conf.set("spark.driver.host", "localhost");
+            conf.set("spark.serializer", "org.apache.spark.serializer.KryoSerializer");
```
marekhorst, line 31: Not sure this is crucial but after we refactor the tests we will be missing the following line:

`conf.setAppName(${someTestClassName}.class.getSimpleName());`

marekhorst (approved):
LGTM even though I had some questions.

iis-common/src/test/java/eu/dnetlib/iis/common/spark/TestWithSharedSparkSession.java
```diff
@@ -10,6 +11,7 @@
 /**
  * Support for tests using {@link SparkSession} - extends this class to access SparkSession in local mode.
  */
+@SlowTest
 public class TestWithSharedSparkSession {
     private static transient SparkSession _spark;
```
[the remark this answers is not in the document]
przemyslawjacewicz: I've taken the general structure of this class from this class [SharedJavaSparkContext](https://github.com/holdenk/spark-testing-base/blob/master/core/src/main/2.0/java/com/holdenkarau/spark/testing/SharedJavaSparkContext.java) thinking that there must be some reason for making `SparkSession` a `transient` variable. But I've tested the code with `SparkSession` as a non-transient variable and tests pass. So maybe making `SparkSession` transient is necessary when testing in some distributed environment. In our test setup this seems not necessary so I think we can safely remove `transient` property from `SparkSession` and I'll do just that :)

iis-common/src/test/java/eu/dnetlib/iis/common/spark/TestWithSharedSparkSession.java
przemyslawjacewicz: Forgot about that, so I'm adding this conf parameter. Funny thing, but when creating the same functionality for `SparkContext` in similar class an error was thrown because application name was missing.

przemyslawjacewicz:
Jenkins build timed out after 1h. The cause is described in issue #1191 and is resolved in PR #1192 . We should first merge #1192 and then merge new master into this PR to resolve long build time.

przemyslawjacewicz:
After pushing code review changes the new build tool 52min so it was successful. I'm merging this PR.

PR #1190, opened 2020-11-12
Closes #1185: Refactor tests with SparkContext to use common base class for SparkContext management

### LINKED ISSUE #1185: Refactor tests with SparkContext to use common base class for SparkContext management
The normal approach for creating `SparkContext` in tests is to create it manually in 'before' methods and close it in 'after' methods. This approach can be converted to more generic one by creating a test base class that takes care of creating and disposing a `SparkContext` instance. `TestWithSharedSparkSession` introduced in PR #1179 does this for `SparkSession`. The goal of this issue is to create such base class for `SparkContext` management and use it in all tests using `SparkContext`.

przemyslawjacewicz:
This PR closes #1185  .

Each test extending `eu.dnetlib.iis.common.spark.TestWithSharedSparkContext` will automatically be tagged as slow test.

One thing to note is that if a test extending `eu.dnetlib.iis.common.spark.TestWithSharedSparkContext` has also `BeforeEach` method then test's `BeforeEach` method must call base class `BeforeEach` method by using `super.beforeEach()`.

### DIFF
--- a/iis-common/src/test/java/eu/dnetlib/iis/common/spark/TestWithSharedSparkContext.java
+++ b/iis-common/src/test/java/eu/dnetlib/iis/common/spark/TestWithSharedSparkContext.java
@@ -1,0 +1,50 @@
+package eu.dnetlib.iis.common.spark;
+
+import eu.dnetlib.iis.common.SlowTest;
+import org.apache.spark.SparkConf;
+import org.apache.spark.SparkContext;
+import org.apache.spark.api.java.JavaSparkContext;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeEach;
+
+import java.util.Objects;
+
+/**
+ * Support for tests using {@link org.apache.spark.SparkContext} - extends this class to access SparkContext in local mode.
+ */
+@SlowTest
+public class TestWithSharedSparkContext {
+    private static transient SparkContext _sc;
+    private static transient JavaSparkContext _jsc;
+    protected boolean initialized = false;
+
+    protected SparkContext sc() {
+        return _sc;
+    }
+
+    protected JavaSparkContext jsc() {
+        return _jsc;
+    }
+
+    @BeforeEach
+    public void beforeEach() {
+        initialized = Objects.nonNull(_sc);
+
+        if (!initialized) {
+            SparkConf conf = new SparkConf();
+            conf.setAppName(getClass().getSimpleName());
+            conf.setMaster("local");
+            conf.set("spark.driver.host", "localhost");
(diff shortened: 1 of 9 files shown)


iis-common/src/test/java/eu/dnetlib/iis/common/spark/TestWithSharedSparkContext.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.Objects;
+
+/**
+ * Support for tests using {@link org.apache.spark.SparkContext} - extends this class to access SparkContext in local mode.
+ */
+@SlowTest
+public class TestWithSharedSparkContext {
+    private static transient SparkContext _sc;
+    private static transient JavaSparkContext _jsc;
```
marekhorst, line 18: Similar question to the one from other PR, out of curiosity: why do we have to declare those two as transient? Is the test class being serialized?

iis-common/src/test/java/eu/dnetlib/iis/common/spark/TestWithSharedSparkContext.java
```diff
@@ -0,0 +1,50 @@
...
+    protected boolean initialized = false;
+
+    protected SparkContext sc() {
+        return _sc;
+    }
```
marekhorst, line 23: Do we really need to have `SparkContext` exposed outside? I didn't find any reference in the changed code. Unless it is exposed for a potential future use...

marekhorst (approved):
LGTM with minor comments.

iis-common/src/test/java/eu/dnetlib/iis/common/spark/TestWithSharedSparkContext.java
przemyslawjacewicz: The same [answer](https://github.com/openaire/iis/pull/1189#discussion_r523007224) applies also here, even more because this PR deals with `SparkContext`, not `SparkSession`.

iis-common/src/test/java/eu/dnetlib/iis/common/spark/TestWithSharedSparkContext.java
przemyslawjacewicz: Again, a blind copy from source project, removing the method.

przemyslawjacewicz:
Jenkins build timed out after 1h. The cause is described in issue #1191 and is resolved in PR #1192 . We should first merge #1192 and then merge new master into this PR to resolve long build time.

przemyslawjacewicz:
After pushing code review changes the new build tool 54min so it was successful. I'm merging this PR.

PR #1197, opened 2020-11-28
Closes #1050: Handle trust level threshold in citation matching exporter module

### LINKED ISSUE #1050: Handle trust level threshold in citation matching exporter module
Currently citation matching outcome, unlike the outcome of all the other mining modules, is exported as BLOB entry (XML record including position, reference text and various identifiers of matched publications) and is not filtered by trust level threshold.

We should add support for `export_trust_level_threshold_document_referencedDocuments` IIS input parameter and filter-out all the internally matched references (typed as `openaire`) with trust level below this threshold. 

Code responsible for building `BlobCitationEntry` is defined in `CitationsActionBuilderModuleUtils` so this is the place we should alter the logic.

przemyslawjacewicz:
This PR adds trust level threshold handling in citation matching exporter module.

Inside this PR
* `eu.dnetlib.iis.wf.export.actionmanager.module.CitationsActionBuilderModuleFactory`changes adding trust level threshold handling plus refactorization
* `eu.dnetlib.iis.wf.export.actionmanager.module.CitationsActionBuilderModuleFactory` unit tests addition and refactorization
* `eu.dnetlib.iis.wf.export.actionmanager.module.CitationsActionBuilderModuleUtils` refactorization and unit test addition

### DIFF
--- a/iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
+++ b/iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
@@ -1,14 +1,5 @@
 package eu.dnetlib.iis.wf.export.actionmanager.module;
 
-import java.util.Collections;
-import java.util.List;
-import java.util.SortedSet;
-import java.util.TreeSet;
-
-import org.apache.commons.collections.CollectionUtils;
-import org.apache.commons.lang.StringUtils;
-import org.apache.hadoop.conf.Configuration;
-
 import eu.dnetlib.dhp.schema.oaf.ExtraInfo;
 import eu.dnetlib.dhp.schema.oaf.Result;
 import eu.dnetlib.iis.common.InfoSpaceConstants;
@@ -18,12 +9,20 @@ import eu.dnetlib.iis.common.model.extrainfo.citations.BlobCitationEntry;
 import eu.dnetlib.iis.common.model.extrainfo.converter.CitationsExtraInfoConverter;
 import eu.dnetlib.iis.export.schemas.Citations;
 import eu.dnetlib.iis.wf.export.actionmanager.cfg.StaticConfigurationProvider;
+import eu.dnetlib.iis.wf.export.actionmanager.entity.ConfidenceLevelUtils;
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang.StringUtils;
+import org.apache.hadoop.conf.Configuration;
+
+import java.util.*;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 /**
  * {@link Citations} based action builder module.
- * 
- * @author mhorst
  *
+ * @author mhorst
  */
 public class CitationsActionBuilderModuleFactory extends AbstractActionBuilderFactory<Citations, Result> {
 
@@ -31,50 +30,50 @@ public class CitationsActionBuilderModuleFactory extends AbstractActionBuilderFa
     private static final String EXTRA_INFO_TYPOLOGY = ExtraInfoConstants.TYPOLOGY_CITATIONS;
 
     // ------------------------ CONSTRUCTORS --------------------------
-    
+
     public CitationsActionBuilderModuleFactory() {
         super(AlgorithmName.document_referencedDocuments);
     }
 
     // ------------------------ LOGIC ---------------------------------
-    
+
     @Override
     public ActionBuilderModule<Citations, Result> instantiate(Configuration config) {
         return new CitationActionBuilderModule(provideTrustLevelThreshold(config));
     }
 
     // ------------------------ INNER CLASS  --------------------------
-    
+
     class CitationActionBuilderModule extends AbstractEntityBuilderModule<Citations, Result> {
 
-        private CitationsExtraInfoConverter converter = new CitationsExtraInfoConverter();
+        private CitationsExtraInfoConverter citationsExtraInfoConverter = new CitationsExtraInfoConverter();
+        private CitationEntriesConverter citationEntriesConverter = new CitationEntriesConverter();
 
         // ------------------------ CONSTRUCTORS --------------------------
-        
+
         /**
          * @param trustLevelThreshold trust level threshold or null when all records should be exported
-         * @param agent action manager agent details
-         * @param actionSetId action set identifier
          */
         public CitationActionBuilderModule(Float trustLevelThreshold) {
             super(trustLevelThreshold, buildInferenceProvenance());
         }
 
         // ------------------------ LOGIC --------------------------
-        
+
         @Override
         protected Class<Result> getResultClass() {
             return Result.class;
         }
-        
+
         @Override
         protected Result convert(Citations source) {
             if (CollectionUtils.isNotEmpty(source.getCitations())) {
                 Result result = new Result();
                 result.setId(source.getDocumentId().toString());
-                
+
                 ExtraInfo extraInfo = new ExtraInfo();
-                extraInfo.setValue(converter.serialize(normalize(source.getCitations())));
+                extraInfo.setValue(citationsExtraInfoConverter.serialize(
+                        citationEntriesConverter.convert(source.getCitations(), getTrustLevelThreshold())));
                 extraInfo.setName(EXTRA_INFO_NAME);
                 extraInfo.setTypology(EXTRA_INFO_TYPOLOGY);
                 extraInfo.setProvenance(this.getInferenceProvenance());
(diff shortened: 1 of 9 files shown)


iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
```diff
@@ -18,101 +9,203 @@
...
+
+        /**
+         * Allows to validate citation entry match.
+         */
+        public static class CitationEntryMatchValidator {
+            private ConfidenceLevelValidator confidenceLevelValidator = new ConfidenceLevelValidator();
+
+            public Boolean validate(CitationEntry citationEntry, Float confidenceLevelThreshold) {
+                if (Objects.isNull(citationEntry.getConfidenceLevel())) {
+                    return false;
```
marekhorst, line 144: This comment is partially applicable to the lines 109-112 but the condition logic I am referring to is part of the code I have marked here.

~~WARN: if I understand this properly this could rule out all the citation entries which were unmatched. We definitely DO NOT want this to happen!~~

I have studied the code more thoroughly and it seems we will not drop the unmatched entries from the list of bibrefs (it is not part of the filter in streaming processing). But still the logic here triggers unnecessary logic of nulling `destinationDocumentId` and `convidenceLevel` (which is already null but this is not the core of the problem) fields and suggests such entry is invalid (as the validator returns `false`) what may give impression there is something wrong with the record which is not the case: it simply does not convey the outcome of citation matching (this is why the confidence level is `null`) but it definitely brings the text and possible external ids we want to make a part of the BLOB.

We should focus on removing matches (or normalizing the matched ids) by applying the `CitationEntryNormalizer` but ONLY on matched entries (indicated by confidence level and destination document id presence). So we should refactor it in a way missing confidence level should not yield the negative outcome of validation process (but the positive outcome should not trigger `validMatchCitationEntryNormalizer`). We could rename the validator to be more adequate to what it is doing and apply norimalizer on matched entries only.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleUtils.java
```diff
@@ -1,56 +1,132 @@
...
+        blobCitationEntry.setIdentifiers(identifiersFn.apply(citationEntry));
+        return blobCitationEntry;
+    }
+
+    /**
+     * Build position value of blob citation entry from citation entry.
+     */
+    public static class PositionBuilder {
+        public static Integer build(CitationEntry citationEntry) {
+            return citationEntry.getPosition();
```
marekhorst, line 55: Do we need a whole function to cover this simple one-liner? I will make a generic comment when summarizing the code review.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleUtils.java
```diff
@@ -1,56 +1,132 @@
...
+                    TypedIdFromDestinationDocumentIdBuilder::build,
+                    TypedIdsFromExternalDestinationDocumentIdsBuilder::isValid,
+                    TypedIdsFromExternalDestinationDocumentIdsBuilder::build);
+        }
+
+        public static List<TypedId> build(CitationEntry citationEntry,
+                                          Function<CitationEntry, Boolean> destinationDocumentIdValidator,
+                                          Function<CitationEntry, TypedId> typedIdFromDestinationDocumentIdFn,
+                                          Function<CitationEntry, Boolean> externalDestinationDocumentIdsValidator,
+                                          Function<CitationEntry, List<TypedId>> typedIdsFromExternalDestinationDocumentIdsFn) {
```
marekhorst, line 84: Again, why so many `Function`s? Do we really need to express the logic this way? It is so difficult to track the code...

marekhorst (changes requested):
OK I have made some comments mainly related to the citation entries which were not a product of citation matching but were conveying the text or external ids only. I think we should refactor this piece of code.

My major concern is: why such excessive usage of `Function`s? It may simplify testing procedure but it makes the code so difficult to read... If the code behind the function is rather simple I would probably stick to the old-fashioned approach. 

Just a generic comment indicating I hardly know what I am reading having major difficulties in tracking the code scattered among different places (and this happens when Functions are everywhere). The requirement for this task was to simply drop the matches if confidence level is below some threshold and we ended up with hundreds of lines of code I can hardly follow.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
```diff
@@ -18,101 +9,203 @@
...
     }
 
     // ------------------------ INNER CLASS  --------------------------
-    
+
     class CitationActionBuilderModule extends AbstractEntityBuilderModule<Citations, Result> {
 
-        private CitationsExtraInfoConverter converter = new CitationsExtraInfoConverter();
+        private CitationsExtraInfoConverter citationsExtraInfoConverter = new CitationsExtraInfoConverter();
+        private CitationEntriesConverter citationEntriesConverter = new CitationEntriesConverter();
```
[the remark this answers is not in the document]
przemyslawjacewicz: Done. I've also renamed field names to match new naming convention.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
przemyslawjacewicz: I should've asked more questions while working on this task :/ 

I've refactored the code that was poorly designed in terms of checking if a citation entry is a result of citation matching. Firstly I've changed the name to 'checker' to show that we only perform a check. Secondly, I've added an enum containing the result of this check. Thirdly, I've refactored the processing flow to validate against threshold and to normalize only citation entries that contain a result of citation matching. 

I've compared the results of running the exporting workflow before and after above changes and I've got the same results, but the code now is much clearer.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleUtils.java
przemyslawjacewicz: The reason behind this was to separate the code and match it to other builders in this class and make tests simpler. This particular code is simple but others are not that simple but to align the code I've also applied this build pattern here. But as discussed offline I'm going to apply this pattern carefully in the future.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleUtils.java
przemyslawjacewicz: Here, the reason was to allow for easy testing - I've wanted to separate the logic to make the test cases smaller and more isolated. As noted in other comment, I'm going to apply this pattern carefully in the future.

przemyslawjacewicz:
I've pushed changes after code review. I've mainly refactored the logic in `CitationsActionBuilderModuleFactory` and changed naming in `ExtraInfoConverter`/`ExtraInfoSerDe` hierarchy.

In terms of usage of `Function` I've wanted to make the code modular, create a proxy when using 'external' classes and make testing easier. But in the future I will use this pattern more carefully. 

Re-requesting code review.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
```diff
@@ -1,118 +1,218 @@
...
+        public SortedSet<BlobCitationEntry> convert(List<CitationEntry> source, Float trustLevelThreshold) {
+            if (source != null) {
+                Float confidenceLevelThreshold = trustLevelConverter.convert(trustLevelThreshold);
+                return source.stream()
+                        .map(citationEntry -> Maps.immutableEntry(citationEntryMatchChecker.check(citationEntry),
+                                citationEntry))
+                        .filter(checkResultAndCitationEntry -> confidenceLevelValidator.validate(
+                                checkResultAndCitationEntry.getKey(), checkResultAndCitationEntry.getValue(), confidenceLevelThreshold))
+                        .map(checkResultAndCitationEntry -> citationEntryNormalizer.normalize(
+                                checkResultAndCitationEntry.getKey(), checkResultAndCitationEntry.getValue()))
```
marekhorst, line 117: OK, so the steps are the following:
* building the `map<k,v>` where the `k` indicates whether citation has matching result or not and `v` conveys the citation entry
* filtering the entries by `confidenceLevel` and removing all the matches below the threshold (!)
* normalizing the destination document id

Do I get it right?

It's quite nice, decouples the steps having dedicated modules realising each functionality. 
The only problem is it does not work as it should because it removes whole citation entries (including the citation text and external identifiers) for matched citations and this not what we wanted to achieve.

Remember `CitationEntry` conveys not only citation matching outcome but also text and external identifiers so basically what the code does is removing such entry from the list. So as a result you will see missing bibref on the list because the text (and possibly external ids) will not be propagated to the XML produced by the exporter module because WHOLE entry is removed.

What we wanted was as I have described during the call: setting `destinationDocumentId` and `confidenceLevel` to `null` for all the matched cases below the threshold. That's all! No filtering, because we don't want to remove any citation from the citation list.

Good thing: you could simplify the whole logic: no need to filter thus no need to generate `CheckResult` and the whole logic could be encoded in a simple map operation `null`ing two fields (when below threshold) I have mentioned above.

The task is relatively simple, this is why I was surprised it takes so much effort and refactoring. Probably there was still some misunderstanding.

Anyway I am closing this review round and will get back once we get this code on track :)

marekhorst (changes requested):
Closing this review round before reaching an end because the logic needs to be a little bit altered.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
przemyslawjacewicz: > OK, so the steps are the following:
> 
>     * building the `map<k,v>` where the `k` indicates whether citation has matching result or not and `v` conveys the citation entry
> 
>     * filtering the entries by `confidenceLevel` and removing all the matches below the threshold (!)
> 
>     * normalizing the destination document id
> 
> 
> Do I get it right?

Yes, you're right. 

> It's quite nice, decouples the steps having dedicated modules realising each functionality.
> The only problem is it does not work as it should because it removes whole citation entries (including the citation text and external identifiers) for matched citations and this not what we wanted to achieve.
> 
> Remember `CitationEntry` conveys not only citation matching outcome but also text and external identifiers so basically what the code does is removing such entry from the list. So as a result you will see missing bibref on the list because the text (and possibly external ids) will not be propagated to the XML produced by the exporter module because WHOLE entry is removed.
> 
> What we wanted was as I have described during the call: setting `destinationDocumentId` and `confidenceLevel` to `null` for all the matched cases below the threshold. That's all! No filtering, because we don't want to remove any citation from the citation list.

I've removed filtering so no citation entries are removed and the logic is encoded using only map operations. Destination document id and confidence level are `null`ed for citation entries that contain the outcome of citation matching and confidence level is below threshold.

przemyslawjacewicz:
I've applied the requested changes and checked the result in Zeppelin using last prod results [here](https://iis-cdh5-test-gw.ocean.icm.edu.pl/zeppelin/#/notebook/2FT37RB19).

Re-requesting code review.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleUtils.java
```diff
@@ -1,56 +1,132 @@
...
+                return build(citationEntry, ConfidenceLevelBuilder::build);
+            }
+
+            public static TypedId build(CitationEntry citationEntry, Function<CitationEntry, Float> confidenceLevelFn) {
+                return new TypedId(citationEntry.getDestinationDocumentId().toString(),
+                        ExtraInfoConstants.CITATION_TYPE_OPENAIRE,
+                        confidenceLevelFn.apply(citationEntry));
+            }
+
+            public static class ConfidenceLevelBuilder {
```
marekhorst, line 110: I guess we should rename it to `TrustLevelBuilder` because it builds trust level in the end...

marekhorst (changes requested):
4 comments, 2 requests for minor fixes, 2 questions :)

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
```diff
@@ -1,118 +1,203 @@
...
+        private CitationEntryNormalizer citationEntryNormalizer = new CitationEntryNormalizer();
+        private BlobCitationEntryBuilder blobCitationEntryBuilder = new BlobCitationEntryBuilder();
+
+        public SortedSet<BlobCitationEntry> convert(List<CitationEntry> source, Float trustLevelThreshold) {
+            if (source != null) {
+                Float confidenceLevelThreshold = trustLevelConverter.convert(trustLevelThreshold);
+                return source.stream()
+                        .map(citationEntry -> confidenceLevelValidator.validate(citationEntry, confidenceLevelThreshold))
+                        .map(citationEntry -> citationEntryNormalizer.normalize(citationEntry))
+                        .map(citationEntry -> blobCitationEntryBuilder.build(citationEntry))
```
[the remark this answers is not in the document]
przemyslawjacewicz: Spark should pack these 3 maps into a single stage, but this class is used in MR framework. I've run the whole export workflow with only citations enabled a few times and it always took about 2/3 minutes.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
```diff
@@ -1,118 +1,203 @@
...
+            private Float conversionFactor;
+
+            public TrustLevelConverter(Float conversionFactor) {
+                this.conversionFactor = conversionFactor;
+            }
+
+            public Float convert(Float trustLevelThreshold) {
+                return Optional.ofNullable(trustLevelThreshold)
+                        .map(x -> x / conversionFactor)
+                        .orElse(null);
```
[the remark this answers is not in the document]
przemyslawjacewicz: IDE does not hint the usage of `Optional` here, it was my initial choice. Your code seems better because I'm unnecessary wrapping the value in an `Optional` and then unwrapping it upon returning. This pattern works better when we have more than one operation to apply to initial value because every operation is isolated. This was probably why I've chosen this solution, but in the future I'll use the simpler solution in situations like this.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
```diff
@@ -1,118 +1,203 @@
...
+
+        /**
+         * Allows to normalize a citation entry.
+         */
+        public static class CitationEntryNormalizer {
+            private CitationEntryMatchChecker citationEntryMatchChecker = new CitationEntryMatchChecker();
+            private MatchingResultCitationEntryNormalizer matchingResultCitationEntryNormalizer = new MatchingResultCitationEntryNormalizer();
+
+            public CitationEntry normalize(CitationEntry citationEntry) {
+                if (citationEntryMatchChecker.isMatchingResult(citationEntry)) {
```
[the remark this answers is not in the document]
przemyslawjacewicz: Argh, this part is an artifact of the previous version with different normalizers for citation entries with and without the result of citation matching. Applying the changes you've requested: normalizing based on the existence of destination document id in citation entry.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleUtils.java
przemyslawjacewicz: This was my initial name also, but this class sets `confidenceLevel` field in `TypedId` class so to avoid any misunderstandings I took the name from the field.

przemyslawjacewicz:
I've applied the changes, re-requesting the review.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleUtils.java
marekhorst: OK, this is another advantage of doing code reviews: finding the bugs in existing code. We should either rename it within the XML blob (within TypedId) or skip applying the multiplier to reduce it to trust level.

I will create a dedicated ticket to handle this, no need to take care of it here.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
```diff
@@ -165,21 +165,20 @@ public CitationEntry validate(CitationEntry citationEntry, Float confidenceLevel
...
         public static class CitationEntryNormalizer {
-            private CitationEntryMatchChecker citationEntryMatchChecker = new CitationEntryMatchChecker();
-            private MatchingResultCitationEntryNormalizer matchingResultCitationEntryNormalizer = new MatchingResultCitationEntryNormalizer();
+            private DestinationDocumentIdNormalizer destinationDocumentIdNormalizer = new DestinationDocumentIdNormalizer();
 
             public CitationEntry normalize(CitationEntry citationEntry) {
-                if (citationEntryMatchChecker.isMatchingResult(citationEntry)) {
-                    return matchingResultCitationEntryNormalizer.normalize(citationEntry);
+                if (Objects.nonNull(citationEntry.getDestinationDocumentId())) {
+                    return destinationDocumentIdNormalizer.normalize(citationEntry);
```
marekhorst, line 172: Thanks but we could make it even more straightforward: since we are checking the `destinationDocumentId` for existence in order to apply a modification on it then I hardly see a point in applying the normalization on the whole `citationEntry` object which is quite confusing. 

I think there is no need to overcomplicate, we could easily skip the `DestinationDocumentIdNormalizer` layer accepting `citationEntry` and replace:

```
destinationDocumentIdNormalizer.normalize(citationEntry);
```
with the content of this unnecessary "wrapping" method:
```
citationEntry.setDestinationDocumentId(destinationDocumentIdNormalizerFn.apply(citationEntry.getDestinationDocumentId()));
```
You check the document id and you apply the modification on it straight away. Doesn't it look simpler?

marekhorst (approved):
Approving, with comments though :)

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/CitationsActionBuilderModuleFactory.java
przemyslawjacewicz: Done :)

przemyslawjacewicz:
One more change pushed. The change is exactly as requested in the comment so I'm not re-requesting the review and I'll merge after successful CI build.

PR #1204, opened 2020-12-18
Closes #1183: Refactor Spark jobs to use SparkSessionSupport for externalization of SparkSession management

### LINKED ISSUE #1183: Refactor Spark jobs to use SparkSessionSupport for externalization of SparkSession management
The normal approach for managing `SparkSession` within Spark job is to close it at the end of the processing. This is most often done using try-with-resource clause. However, this causes problems in job tests when we want to use `SparkSession` created in test after job execution. Some jobs and their tests use an approach of externalizing `SparkSession` management and having it managed by the test class, not the job class. This is done by supplying a boolean flag indicating if the `SparkSession` is managed by the job. The goal of this issue is to align this approach to use `SparkSessionSupport#runWithSparkSession` method (introduced in PR #1179) in all tests and jobs using externalization of `SparkSession` management.

przemyslawjacewicz:
Refactors Spark jobs to use `SparkSessionSupport` for running jobs that share `SparkSession` with tests.

I've changed the way we mark `SparkSession` to be not stopped after job finished. `SparkSessionSupport` used a flag `isSparkSessionManaged` and when the flag was set to `true` the session was stopped after running user code. So to not stop the session user had to set the flag to `false`. I've changed this to the way it is done in `eu.dnetlib.iis.wf.importer.infospace.ImportInformationSpaceJob` where an 'opposite' flag was used, that is when the value of the flag is `true` the session is not closed. This allows a more convenient usage of the flag because we only have to pass the flag as job argument and this will set the flag to `true`. We do not have to provide any explicit value for the flag. The default value of `false` means that the session will be stopped at the end. This refactoring forced a name change for the flag so I've used a more flag-oriented name of `sharedSparkSession`.

### DIFF
--- a/iis-common/src/main/java/eu/dnetlib/iis/common/spark/SparkSessionSupport.java
+++ b/iis-common/src/main/java/eu/dnetlib/iis/common/spark/SparkSessionSupport.java
@@ -41,7 +41,7 @@ public class SparkSessionSupport {
      * when SparkSession is shared e.g. created in tests. Allows to reuse SparkSession created externally.
      *
      * @param conf               SparkConf instance
-     * @param sharedSparkSession When true will stop SparkSession
+     * @param sharedSparkSession When true will not stop SparkSession
      * @param job                Job using constructed SparkSession
      */
     public static void runWithSparkSession(Function<SparkConf, SparkSession> sparkSessionBuilder,
--- a/iis-common/src/main/java/eu/dnetlib/iis/common/spark/SparkSessionSupport.java
+++ b/iis-common/src/main/java/eu/dnetlib/iis/common/spark/SparkSessionSupport.java
@@ -20,49 +20,49 @@ public class SparkSessionSupport {
 
     /**
      * Runs a given job using SparkSession created using default builder and supplied SparkConf. Stops SparkSession
-     * when SparkSession is managed. Allows to reuse SparkSession created externally.
+     * when SparkSession is shared e.g. created in tests. Allows to reuse SparkSession created externally.
      *
-     * @param conf                  SparkConf instance
-     * @param isSparkSessionManaged When true will stop SparkSession
-     * @param job                   Job using constructed SparkSession
+     * @param conf               SparkConf instance
+     * @param sharedSparkSession When true will not stop SparkSession
+     * @param job                Job using constructed SparkSession
      */
     public static void runWithSparkSession(SparkConf conf,
-                                           Boolean isSparkSessionManaged,
+                                           Boolean sharedSparkSession,
                                            Job job) {
         runWithSparkSession(
                 SparkSessionFactory::withConfAndKryo,
                 conf,
-                isSparkSessionManaged,
+                sharedSparkSession,
                 job);
     }
 
     /**
      * Runs a given job using SparkSession created using supplied builder and supplied SparkConf. Stops SparkSession
-     * when SparkSession is managed. Allows to reuse SparkSession created externally.
+     * when SparkSession is shared e.g. created in tests. Allows to reuse SparkSession created externally.
      *
-     * @param conf                  SparkConf instance
-     * @param isSparkSessionManaged When true will stop SparkSession
-     * @param job                   Job using constructed SparkSession
+     * @param conf               SparkConf instance
+     * @param sharedSparkSession When true will stop SparkSession
+     * @param job                Job using constructed SparkSession
      */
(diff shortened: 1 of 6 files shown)


iis-common/src/main/java/eu/dnetlib/iis/common/spark/SparkSessionSupport.java
```diff
@@ -20,49 +20,49 @@ private SparkSessionSupport() {
...
     /**
      * Runs a given job using SparkSession created using default builder and supplied SparkConf. Stops SparkSession
-     * when SparkSession is managed. Allows to reuse SparkSession created externally.
+     * when SparkSession is shared e.g. created in tests. Allows to reuse SparkSession created externally.
      *
-     * @param conf                  SparkConf instance
-     * @param isSparkSessionManaged When true will stop SparkSession
-     * @param job                   Job using constructed SparkSession
+     * @param conf               SparkConf instance
+     * @param sharedSparkSession When true will not stop SparkSession
```
marekhorst, line 26: Could we follow the convention and rename it to `isSparkSessionShared` or something similar?

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/relation/citation/CitationRelationExporterJob.java
```diff
@@ -54,8 +54,8 @@ public static void main(String[] args) {
 
     @Parameters(separators = "=")
     private static class JobParameters {
-        @Parameter(names = "-isSparkSessionManaged", arity = 1)
-        private Boolean isSparkSessionManaged = Boolean.TRUE;
+        @Parameter(names = "-sharedSparkSession")
```
marekhorst, line 57: If we agree on changing the variable name then we should adjust the parameter name in the same way.

marekhorst (approved):
Approving with simple rename suggestion.

przemyslawjacewicz:
I've chosen the name for the flag marking `SparkSession` as shared session to match the convention for naming flags and the usage of booleans in JCommander. So for example in `ImportInformationSpaceJob` we have now the argument defined as 
```java
        @Parameter(names = "-sharedSparkSession")
        private Boolean sharedSparkSession = Boolean.FALSE;
```
This means that when we do not provide the argument as job argument the default value of `false` will be used. Only by explicit usage of the argument we set it to `true`. This is done in `ImportInformationSpaceJobTest`:
```java
        ImportInformationSpaceJob.main(new String[]{
                "-sharedSparkSession",
                "-skipDeletedByInference", Boolean.TRUE.toString(),
                ...
```
This indicates that the `SparkSession` is shared in a flag-way.

Using the name `isSparkSessionShared` would be a bit strange I think. If the argument would be named as `isSparkSessionShared`, in the test class we would have something like this
```java
        ImportInformationSpaceJob.main(new String[]{
                "-isSparkSessionShared",
                "-skipDeletedByInference", Boolean.TRUE.toString(),
                ...
```
I think this would confuse the user because it would seem that the value for `isSparkSessionShared` is missing. So a more natural way now would be to provide the value explicitly
```java
        ImportInformationSpaceJob.main(new String[]{
                "-isSparkSessionShared", "true"
                "-skipDeletedByInference", Boolean.TRUE.toString(),
                ...
```
but this would in turn force a change in class defining the arguments
```java
        @Parameter(names = "-isSparkSessionShared", arity = 1)
        private Boolean isSparkSessionShared = Boolean.FALSE;
```
because JCommander treats booleans as flags by default. I think this is also confusing. 

I think we could consider another approach - changing the name of the field and not changing the name of the argument
```java
        @Parameter(names = "-sharedSparkSession")
        private Boolean isSparkSessionShared = Boolean.FALSE;
```
This would give a flag-oriented usage as job argument in tests 
```java
        ImportInformationSpaceJob.main(new String[]{
                "-sharedSparkSession",
                "-skipDeletedByInference", Boolean.TRUE.toString(),
                ...
```
and conform to java conventions for naming booleans. But this would break the match between argument and field name. Which solution do you think is the best ?

marekhorst:
We have discussed this during today's daily call and the idea was to leave it as it is now. But when I looked into it once again I think we could take advantage of the naming flexibility and go with `sharedSparkSession` as param name and `isSparkSessionShared` as variable name - just as you have proposed in the comment. This would make the java code more readable and also the parameter name would be straightforward.

przemyslawjacewicz:
Ok, I've changed the names, will merge when CI passes.

PR #1205, opened 2020-12-21
Closes #1198: Create a common functionality for converting between confidence level and trust level

### LINKED ISSUE #1198: Create a common functionality for converting between confidence level and trust level
Confidence level and trust level are used within difference contexts. Confidence level is an IIS concept and it's value is between 0 and 1. Trust level is an Information Space concept and is between 0 and 0.9. To convert between the concepts we have to either multiply or divide by scaling factor of 0.9. Currently this conversion is done 'by hand' that is explicitly multiplying or dividing by scaling factor. We should introduce a common functionality for conversion with default scaling factor.

przemyslawjacewicz:
This PR adds support for conversion in both directions between confidence and trust level. 

One thing to look closer into is the scope of refactorization of tests involving conversion between levels. I did not change the tests so they still relay on value comparison using `CONFIDENCE_TO_TRUST_LEVEL_FACTOR` constant. We could refactor the tests to use mock conversion class and in test check if the mock was used.


iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/AbstractBuilderModule.java
```diff
@@ -70,7 +68,7 @@ protected DataInfo buildInference(float confidenceLevel) throws TrustLevelThresh
      * @throws TrustLevelThresholdExceededException thrown when trust level threshold was exceeded
      */
     protected DataInfo buildInference(float confidenceLevel, String inferenceProvenance) throws TrustLevelThresholdExceededException {
-        float currentTrustLevel = confidenceLevel * InfoSpaceConstants.CONFIDENCE_TO_TRUST_LEVEL_FACTOR;
+        float currentTrustLevel = new ConfidenceLevelConverter().convertToTrustLevel(confidenceLevel);
```
marekhorst, line 71: I'd go with static method here and in all similar cases. 

If we agree to rely on a static conversion factor we won't need to instantiate  the immutable object for such a simple task.

marekhorst (changes requested):
I guess we could simplify the transformation logic which in fact is pretty static.

przemyslawjacewicz:
I'm adding a single comment here to summarize the changes.

I've removed two distinct classes and added a single utility class with static methods to convert confidence level to trust level and backwards. Conversion factors are static based on `InfoSpaceConstants.CONFIDENCE_TO_TRUST_LEVEL_FACTOR`. I've changed the names of methods slightly, which not are `confidenceLevelToTrustLevel()` and `trustLevelToConfidenceLevel()` (I think are a bit clearer).

I've also added the versions of conversion methods using primitive `float` as method parameter and method result type. 
```java 
    public static float confidenceLevelToTrustLevel(float confidenceLevel)
    public static float trustLevelToConfidenceLevel(float trustLevel)
```
This was done to avoid NPE warnings when we use primitive type as the result type of conversion. This can be seen in `eu.dnetlib.iis.wf.export.actionmanager.module.AbstractBuilderModule`. The method `buildInference(float confidenceLevel, String inferenceProvenance)` is now using primitive version of conversion method:
```java
    protected DataInfo buildInference(float confidenceLevel, String inferenceProvenance) throws TrustLevelThresholdExceededException {
        float currentTrustLevel = ConfidenceAndTrustLevelConversionUtils.confidenceLevelToTrustLevel(confidenceLevel);
        if (trustLevelThreshold == null || currentTrustLevel >= trustLevelThreshold) {
            if (StringUtils.isNotBlank(inferenceProvenance)) {
                return buildInferenceForTrustLevel(BuilderModuleHelper.getDecimalFormat().format(currentTrustLevel), 
                        inferenceProvenance);
            } else {
                return buildInferenceForTrustLevel(BuilderModuleHelper.getDecimalFormat().format(currentTrustLevel));    
            }
        } else {
            throw new TrustLevelThresholdExceededException();
        }
    }
```
If for conversion we used the method with `Float` parameter then this warning is showed by the compiler
```text
Unboxing of 'ConfidenceAndTrustLevelConversionUtils.confidenceLevelToTrustLevel(confidenceLevel)' may produce 'NullPointerException' 
```
I don't know if this is the best solution here, thought. Re-requesting code review.

iis-common/src/main/java/eu/dnetlib/iis/common/model/conversion/ConfidenceAndTrustLevelConversionUtils.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Converts confidence level to trust level using confidence to trust level scaling factor based on
+     * {@link InfoSpaceConstants#CONFIDENCE_TO_TRUST_LEVEL_FACTOR}. Provides null safety.
+     *
+     * @param confidenceLevel Confidence level to be converted.
+     * @return Float with corresponding trust level value.
+     */
+    public static Float confidenceLevelToTrustLevel(Float confidenceLevel) {
+        return Objects.nonNull(confidenceLevel) ? confidenceLevel * confidenceToTrustConversionFactor : null;
```
marekhorst, line 25: I think a good rule of thumb in such cases, regardless the simplicity of the logic, is to avoid duplicating the code and relying on simple call to the `confidenceLevelToTrustLevel()` method accepting primitives. This would make future refactoring also much simpler.

iis-common/src/main/java/eu/dnetlib/iis/common/model/conversion/ConfidenceAndTrustLevelConversionUtils.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Converts trust level to confidence level using trust to confidence level scaling factor based on
+     * {@link InfoSpaceConstants#CONFIDENCE_TO_TRUST_LEVEL_FACTOR}. Provides null safety.
+     *
+     * @param trustLevel Trust level to be converted.
+     * @return Float corresponding to confidence level value.
+     */
+    public static Float trustLevelToConfidenceLevel(Float trustLevel) {
+        return Objects.nonNull(trustLevel) ? trustLevel * trustToConfidenceConversionFactor : null;
```
marekhorst, line 48: We could rely on `trustLevelToConfidenceLevel()` accepting primitives here to avoid duplicating the logic encoded in this method.

marekhorst (approved):
Approving, although I would recommend minor changes in the code suggested in the comments.

przemyslawjacewicz:
I've changed the conversion to use primitive methods, will merge when CI passes.

PR #1211, opened 2021-01-17
Closes #1178: Consider checking the output directory validity in a prior stage

### LINKED ISSUE #1178: Consider checking the output directory validity in a prior stage
In several IIS mining rounds the reason for a failure was either missing or invalid (in terms of access rights) output directory provided as `output_remote_location` input parameter. Usually it points either to `/var/lib/dnet/actionManager_BETA` or `/var/lib/dnet/actionManager_PROD` and the directory owner has to match also the user executing the IIS job which was not always the case.

We cannot solve this problem within the scope of IIS execution but we could definitely report this problem earlier than at the exporting phase. Currently when the job fails, after 1 day of execution or sometimes more, the DNet workflow also fails and several operations (after fixing the output directory issue and copying the mining outcome from command line) have to be performed manually, e.g. including updating the action set identifiers, which is pretty cumbersome, prone to errors and time consuming.

If we could report such errors (by failing the job with proper error message) just after the IIS execution starts then we could save a lot of time because the output directory could be fixed straight away and the IIS job could be rerun.

przemyslawjacewicz:
This PR adds output location validation for `primary_main` workflow. It is done using oozie's `fs` action with `touchz` command. This command creates an empty file under user specified path using current time as file's creation time. If the file cannot be created an error is raised which marks the output location as not valid. This behavior is easily testable so I've also added a workflow test. 

One thing we might consider is whether to remove the created file - currently it is removed but we could leave it as a marker of workflow execution start.

### DIFF
--- a/iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/main/output-validation/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/main/output-validation/oozie_app/workflow.xml
@@ -1,0 +1,90 @@
+<workflow-app xmlns="uri:oozie:workflow:0.4" name="test-primary_main_output_validation">
+
+    <global>
+        <job-tracker>${jobTracker}</job-tracker>
+        <name-node>${nameNode}</name-node>
+        <configuration>
+            <property>
+                <name>mapreduce.job.queuename</name>
+                <value>${queueName}</value>
+            </property>
+        </configuration>
+    </global>
+
+    <start to="primary_main"/>
+
+    <action name="primary_main">
+        <sub-workflow>
+            <app-path>${wf:appPath()}/primary_main</app-path>
+            <propagate-configuration/>
+            <configuration>
+                <property>
+                    <name>import_islookup_service_location</name>
+                    <value>$UNDEFINED$</value>
+                </property>
+                <property>
+                    <name>import_infospace_graph_location</name>
+                    <value>$UNDEFINED$</value>
+                </property>
+                <property>
+                    <name>import_patent_tsv</name>
+                    <value>$UNDEFINED$</value>
+                </property>
+                <property>
+                    <name>ingest_pmc_cache_location</name>
+                    <value>$UNDEFINED$</value>
+                </property>
+                <property>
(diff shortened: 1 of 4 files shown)


iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/main/output-validation/oozie_app/workflow.xml
```diff
@@ -0,0 +1,90 @@
...
+                    <name>citationmatchingFuzzyNumberOfPartitions</name>
+                    <value>$UNDEFINED$</value>
+                </property>
+                <property>
+                    <name>export_patent_date_of_collection</name>
+                    <value>$UNDEFINED$</value>
+                </property>
+                <property>
+                    <name>output_remote_location</name>
+                    <value>/user/root/</value>
```
marekhorst, line 59: This would work for sure but I am wondering whether we could somehow make sure we have a full control over testing environment.

What I mean is once the `/user/root/` directory is missing (due to whatever reason, e.g. running tests in a test cluster without this particular directory) our test might give unexpected results. 

I guess it will be better if we could create such directory for testing purposes and e.g. apply `chmod` of `<fs>` to restrict the privileges (e.g. by disallowing write access). We would need to drop this directory after the test execution to make sure the next iteration will not fail when cleaning up the env.

marekhorst (changes requested):
Just a single change request. If possible but I guess we could rely on `chmod` or something similar when creating output directory.

iis-wf/iis-wf-primary/src/test/resources/eu/dnetlib/iis/wf/primary/main/output-validation/oozie_app/workflow.xml
przemyslawjacewicz: Argh, you are right, I've been running tests on main cluster where this dir is present, but ci cluster does not have `root` user and therefore `/user/root` does not exist. I'm changing this into temporary dir created within init action with write permissions removed just as you commented.

przemyslawjacewicz:
Changes done, re-requesting code review.

marekhorst (approved):
Looks good, thanks!

PR #1225, opened 2021-02-02
Closes #1213: Align IIS with the recent updates in Oaf model related to OpenAccess status

przemyslawjacewicz:
NOTE: IIS will fail to build after merging this PR because this PR requires the new version of `dhp-schemas` lib, which is updated in #1226 .

PR #1224, opened 2021-02-02
Closes #1212: Align IIS with the recent updates in Oaf model related to PID provenance

przemyslawjacewicz:
NOTE: IIS will fail to build after merging this PR because this PR requires the new version of `dhp-schemas` lib, which is updated in #1226 .

PR #1226, opened 2021-02-02
Closes #1217: Fix dependency issue causing IIS build failure

### LINKED ISSUE #1217: Fix dependency issue causing IIS build failure
The `dhp-schemas:1.2.4-dhp_oaf_model`, which is the new version of one of our DNet dependencies, has caused our master branch build to fail:

http://ci.ceon.pl/view/IIS/job/iis_build_multibranch/job/master/

The module was released with `dhp_oaf_model` suffix to indicate this is not a final release but experimental one. Since DNet dependencies are not defined explicitly in IIS but with version ranges this makes this version to be picked up by our master branch.

This is the original redmine thread with reasoning behind DNet module release:
https://issue.openaire.research-infrastructures.eu/issues/6281#note-10

We should:
1) fix IIS build and integration tests since it is broken for more than a week 
2) find a bulletproof solution for handling such cases in future: when some DNet lib is released with suffix in its version our IIS master branch should not pick this up (if we need to rely on this version we can define it explicitly on a branch we are working on). This may require @claudioatzori involvement in agreeing on the way DNet releases are marked when providing experimental feature (using `qualifier` or some other more appropriate approach).

przemyslawjacewicz:
This PR adds classifier for `dhp-schemas` dependency. This is probably a temporary solution which should be reverted when new `dhp-schemas` is released without any classifiers.

NOTE: IIS will fail to build after merging this PR because the new version requires code changes which are done in PR #1224 and in #1225 .

### DIFF
--- a/iis-common/pom.xml
+++ b/iis-common/pom.xml
@@ -35,6 +35,7 @@
         <dependency>
             <groupId>eu.dnetlib.dhp</groupId>
             <artifactId>dhp-schemas</artifactId>
+            <classifier>dhp_oaf_model</classifier>
         </dependency>
 
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -495,6 +495,7 @@
                 <groupId>eu.dnetlib.dhp</groupId>
                 <artifactId>dhp-schemas</artifactId>
                 <version>[1.0.0, 2.0.0)</version>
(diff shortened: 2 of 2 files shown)


marekhorst (approved):
LGTM.

PR #1229, opened 2021-02-08
Closes #1199: Align on representing confidence level against exported citations encoded as XML

### LINKED ISSUE #1199: Align on representing confidence level against exported citations encoded as XML
While reviewing the code for #1197 we've discovered there is a misalignment in a way we represent confidence against the matched citation entries. 

The field in exported XML BLOB is named `confidenceLevel` but in fact we apply the `0.9` factor to normalize the confidence level value (0-1 values range) into the trust level (0-0.9 values range) which is the common representation of trust within the information space.

We should either rename the field to `trustLevel` or simply drop the downgrading the value by the `0.9` factor. Just need to agree with @claudioatzori and Argiro on what suits better here (probably sticking to confidence level field name).

Original review note: https://github.com/openaire/iis/pull/1197#discussion_r538322786

przemyslawjacewicz:


marekhorst (approved):
LGTM.

Issue #1238, opened 2021-02-25
Remove unnecessary filtering of blank cache records in webcrawler and patent metadata retriever


przemyslawjacewicz:
Ticket #1221 will remove writing of empty records to webcrawler and patent metadata retriever caches. We should also remove already existing empty records in caches to avoid their propagation to next caches. The removal is implemented as a filter on RDD read from cache. The filter will be unnecessary after first IIS run after closing #1221 so we should remove it.

Issue #1234, opened 2021-02-11
Multiple requests when retrieving content from external services


przemyslawjacewicz:
Running cached webcrawler job for issue #1136 revealed that we're retrieving content from external services multiple times. For example running the job with only a single site to read results in three requests. We should send only one request for each site to read. This looks like a Spark caching problem, that is we're not caching the responses from the services and when another Spark action is called with the same RDD we're executing the whole plan again.

We should fix this for cached webcrawler and check if the same situation is for patent metadata retriever.

przemyslawjacewicz:
I'm adding yarn logs dump from cached webcrawler run with single site to read (to avoid log removal): [application_1603828052984_40087.log](https://github.com/openaire/iis/files/5966116/application_1603828052984_40087.log) .

przemyslawjacewicz:
We should implement a mechanism in local job tests to check how many times a content is requested. This can be done by checking how many times a content is served in test factories and fail if a content is requested more than one time.

PR #1239, opened 2021-02-26
Closes #1221: Remove empty entries creation in output and cache

### LINKED ISSUE #1221: Remove empty entries creation in output and cache
Current implementation of `CachedWebCrawlerJob` and `PatentMetadataRetrieverJob` creates an 'empty' entry in `text` cache and in output for each failed response from a service. This implementation is not usual - a response should be either a successful response and be stored in output and in `text` cache OR be a failed response and be stored in fault and in `fault` cache. We should remove creation of an empty entry in output and in `text` cache for each service response with a failure. This will make the code clearer and also make testing less painful. To do this we must check if any downstream action is dependent on an empty entry in output and in `text` cache.

przemyslawjacewicz:
The main changes are in classes `PatentMetadataRetrieverJob` and `CachedWebCrawlerJob` so reviewing changes should start from there. There are a few minor additional changes like unification of some logic in both classes and some reorganization in test classes. 

Two `TODO` statements are committed, both concern removing of filtering after cache migration, that is task #1238.


iis-wf/iis-wf-referenceextraction/src/test/resources/eu/dnetlib/iis/wf/referenceextraction/patent/data/retriever/report.json
```diff
@@ -1,3 +1,3 @@
 {"key":"processing.referenceExtraction.patent.retrieval.fromCache.total", "type":"COUNTER", "value": "0"}
-{"key":"processing.referenceExtraction.patent.retrieval.processed.total", "type":"COUNTER", "value": "2"}
+{"key":"processing.referenceExtraction.patent.retrieval.processed.total", "type":"COUNTER", "value": "1"}
```
marekhorst, line 2: Hmm, not so sure about this.
If we had one successfully processed record and one resulting in failure I would expect to have `2` among `processed.total`. Fixing the way we store cached entries (by removing empty entry) should not affect the counters and the state we expose to clients.

iis-wf/iis-wf-referenceextraction/src/test/resources/eu/dnetlib/iis/wf/referenceextraction/patent/data/retriever/report_from_cache.json
```diff
@@ -1,3 +1,3 @@
-{"key":"processing.referenceExtraction.patent.retrieval.fromCache.total", "type":"COUNTER", "value": "2"}
+{"key":"processing.referenceExtraction.patent.retrieval.fromCache.total", "type":"COUNTER", "value": "1"}
```
marekhorst, line 1: I think the similar remark is applicable here: if we had one successfully processed record and one failure stored in cache then I would expect to see `2` among `fromCache.total` because in fact we retrieved information about 2 records from cache. 

This case is definitely more debatable than the first one so we can discuss what would be more meaningful. Although again, this should not be changed only because we stopped storing empty records in cache. We could change the semantics behind `fromCache.total` but independently from the core change of this pull request.

iis-wf/iis-wf-referenceextraction/src/test/resources/eu/dnetlib/iis/wf/referenceextraction/softwareurl/data/webcrawler/nosource/report.json
```diff
@@ -1,3 +1,3 @@
 {"key":"processing.referenceExtraction.softwareUrl.webcrawl.fromCache.total", "type":"COUNTER", "value": "0"}
-{"key":"processing.referenceExtraction.softwareUrl.webcrawl.processed.total", "type":"COUNTER", "value": "2"}
+{"key":"processing.referenceExtraction.softwareUrl.webcrawl.processed.total", "type":"COUNTER", "value": "0"}
```
marekhorst, line 2: Was this change deliberate or was it introduced just because the module started to return `0` instead of `2`? ;)

This is one of the 3 comments about the same issue: `processed.total`, at least in my opinion, should sum both successfully processed records and `processed.fault`.

marekhorst (changes requested):
I think there is one issue which should be discussed.

przemyslawjacewicz:
I'm leaving a single response here because it applies to all comments. I did not think over report logic and make changes to update it after removing empty records from cache and output. So you're right - the changes in test jsons reflected what started to be returned after changes ;) 

I've restored the original meaning of report counters for `processed.total` and for `from.cache` counters. This required a few additional RDD operations. Re-requesting code review.

iis-wf/iis-wf-referenceextraction/src/main/java/eu/dnetlib/iis/wf/referenceextraction/patent/PatentMetadataRetrieverJob.java
```diff
@@ -137,9 +137,16 @@ public static void main(String[] args) throws Exception {
             }
 
             // store final results
+            long entitiesReturnedFromCacheCount = entitiesReturnedFromCache.count();
+            long faultsReturnedFromCacheCount = inputJoinedWithCache
+                    .filter(x -> x._2._2.isPresent() && !x._2._2.get().isPresent())
+                    .count();
+            long processedEntitiesCount = retrievedPatentMeta.count();
+            long processedFaultsCount = faults.count();
```
marekhorst, line 145: I guess we should track all the RDDs we are running `count()` method on. If they are also persisted in a later stage (e.g. `faults`) or are part of other operations which end up with an action (`retrievedPatentMeta` probably is the case) then they are potential subject for caching.

iis-wf/iis-wf-referenceextraction/src/main/java/eu/dnetlib/iis/wf/referenceextraction/softwareurl/CachedWebCrawlerJob.java
```diff
@@ -153,12 +153,21 @@ public static void main(String[] args) throws Exception {
...
-            
+
+            long entitiesReturnedFromCacheCount = entitiesReturnedFromCache.count();
+            long faultsReturnedFromCacheCount = inputJoinedWithCache
+                    .filter(x -> x._2._2.isPresent() && !x._2._2.get().isPresent())
+                    .count();
+            long processedEntitiesCount = webcrawledEntities.count();
+            long processedFaultsCount = faults.mapToPair(e -> new Tuple2<>(e.getInputObjectId(), 1))
+                    .join(inputByUrl)
+                    .count();
```
marekhorst, line 164: I guess the count related comment could be applicable here as well. We should check whether it does induce the caching requirement.

marekhorst (approved):
Just two remarks related to potential requirement of using caches. Please look into this - if you think the RDD caching is not needed then just merge it.

przemyslawjacewicz:
> Just two remarks related to potential requirement of using caches. Please look into this - if you think the RDD caching is not needed then just merge it.

Yes, you're definitely right that we should check for any potential for caching. As discussed offline we have issue #1234 in line which is probably also connected with caching so I thought we should do it together there. So I'm merging this PR and leaving caching changes to #1234.

PR #1240, opened 2021-03-03
Closes #1221: Remove empty entries creation in output and cache - FIX

### LINKED ISSUE #1221: Remove empty entries creation in output and cache

przemyslawjacewicz:
Fixing failing integration test suite.

After removing empty records from patent metadata retriever output patent metadata extractor input changes. Instead of 7 input records (1 with non-empty text and 6 with empty text) we have now only 1 input record with non-empty text. This PR fixes report entries to match these changes.

### DIFF
--- a/iis-wf/iis-wf-referenceextraction/src/test/resources/eu/dnetlib/iis/wf/referenceextraction/patent/data/report_patent_metadata_extraction.json
+++ b/iis-wf/iis-wf-referenceextraction/src/test/resources/eu/dnetlib/iis/wf/referenceextraction/patent/data/report_patent_metadata_extraction.json
@@ -1,2 +1,2 @@
-{"key":"processing.referenceExtraction.patent.metadataextraction.processed.total", "type":"COUNTER", "value": "7"}
-{"key":"processing.referenceExtraction.patent.metadataextraction.processed.fault", "type":"COUNTER", "value": "6"}
+{"key":"processing.referenceExtraction.patent.metadataextraction.processed.total", "type":"COUNTER", "value": "1"}
+{"key":"processing.referenceExtraction.patent.metadataextraction.processed.fault", "type":"COUNTER", "value": "0"}


iis-wf/iis-wf-referenceextraction/src/test/resources/eu/dnetlib/iis/wf/referenceextraction/patent/data/report_patent_metadata_extraction.json
```diff
@@ -1,2 +1,2 @@
-{"key":"processing.referenceExtraction.patent.metadataextraction.processed.total", "type":"COUNTER", "value": "7"}
-{"key":"processing.referenceExtraction.patent.metadataextraction.processed.fault", "type":"COUNTER", "value": "6"}
\ No newline at end of file
+{"key":"processing.referenceExtraction.patent.metadataextraction.processed.total", "type":"COUNTER", "value": "1"}
+{"key":"processing.referenceExtraction.patent.metadataextraction.processed.fault", "type":"COUNTER", "value": "0"}
```
marekhorst, line 2: Does it mean the change related only to removing empty entries in cache and output also affects the number of faults generated by patent metadata extraction process? 

I mean it is unexpected, to me, but if it is justifiable then let's leave it as it is.

marekhorst (commented):
Just wanted to clarify on why we have this drop in number of faults before we merge this change.

Let's treat those failed tests as fuses :)

iis-wf/iis-wf-referenceextraction/src/test/resources/eu/dnetlib/iis/wf/referenceextraction/patent/data/report_patent_metadata_extraction.json
przemyslawjacewicz: Yes, the number of faults generated by extractor drops because we do not propagate faults from retriever as empty records to extractor. 

Before removing empty records from output we had 7 input records to retriever for which we had 7 records as output (1 non-empty and 6 empty) and 6 records as faults. The output was propagated as input to extractor. So extractor processed 7 input records with only 1 non-empty. Out of those 7 records only 1 was successfully processed and this was reflected in the report. 

After removing empty records from output we have 7 input records to retriever for which we have 1 record as output and 6 records as faults. The one output record is propagated as input to extractor which successfully processes it. And this is reflected in the new values in extractor report.

przemyslawjacewicz:
I think the drop in number of faults is justifiable because we do not propagate faults from retriever to extractor. Re-requesting review.

marekhorst (approved):
OK, then it sounds it was expected in the end. Thanks for explanation, got lost between different modules :)

PR #1257, opened 2021-04-07
Closes #1231: Refactor HDFS file counting in tests to use dedicated method accepting extension 

### LINKED ISSUE #1231: Refactor HDFS file counting in tests to use dedicated method accepting extension
File counting in tests is done using `HdfsUtils.countFiles` method with an explicit path filter. Most of the times we count files with a specific extension. We should use dedicated method `HdfsUtils.countFiles` accepting file extension for this use case.

przemyslawjacewicz:


marekhorst (approved):
LGTM.

PR #1258, opened 2021-04-07
Closes #1209: CitationRelationExporterJob uses wrong name for evaluated confidence level

### LINKED ISSUE #1209: CitationRelationExporterJob uses wrong name for evaluated confidence level
`CitationRelationExporterJob` uses `ConfidenceLevelUtils#evaluateConfidenceLevelThreshold` to evaluate and convert input trust level threshold value to confidence level threshold. However,  the variable name is `trustLevelThreshold` which is wrong. It should be changed to `confidenceLevelThreshold`.

przemyslawjacewicz:


marekhorst (approved):
OK, so it does not introduce any functional changes, right? It's just a rename of the `Float` field.

Anyway, LGTM.

przemyslawjacewicz:
> OK, so it does not introduce any functional changes, right? It's just a rename of the `Float` field.
> 
> Anyway, LGTM.

Yes, it's just a trivial name change. I've found it when working on confidence/trust level converters.

PR #1259, opened 2021-04-08
Closes #1201: Refactor Spark support for avro datastores

### LINKED ISSUE #1201: Refactor Spark support for avro datastores
We have two classes supporting avro datastores in Spark
* `eu.dnetlib.iis.common.spark.avro.AvroDataFrameSupport`
* `eu.dnetlib.iis.common.spark.avro.AvroDatasetSupport`

Both classes support reading and writing of avro datastores with Spark. They also support other Spark related operations that is creation of dataframes with avro types and conversion between dataframes and datasets. We should consider splitting both classes into classes responsible for IO and responsible for other Spark operations. IO related classes could be named as `AvroDataFrameStorage` and `AvroDatasetStorage`. Current name could be reused for other Spark operations.

przemyslawjacewicz:


marekhorst (approved):
LGTM.

PR #1242, opened 2021-03-09
Rename RCUK->UKRI and GSRT->GSRI

johnfouf:
Rename RCUK and GSRT funders

PR #1262, opened 2021-04-12
Closes #1250: Rename RCUK and GSRT projects

### LINKED ISSUE #1250: Rename RCUK and GSRT projects
We should integrate this pull request:
https://github.com/openaire/iis/pull/1242

renaming:
- RCUK->UKRI
- GSRT->GSRI

This does not have to go through the data curation process because the change is rather trivial. We should just align our integration tests to reflect new projects naming.

przemyslawjacewicz:
This PR integrates #1242  and #1261 with updates in integration tests.

The updates in integration tests are 'artificial' that is, are not based on real data from PROD or BETA. It's because the projects are not fully migrated to new funders. As a result test cases are based on previous version with modified funding class names. I've also modified project ids to match new funder name.


marekhorst (approved):
LGTM.

Issue #1252, opened 2021-03-19
Align IIS with the recent updates in Oaf model related to the introduction of instance.alternateIdentifiers


marekhorst:
More details on the changes in Oaf model are available here:
https://issue.openaire.research-infrastructures.eu/issues/6487

przemyslawjacewicz:
Updates merged into temporary branch przemyslawjacewicz_1251_1252_alignment_with_updates_in_oaf_model with this PR #1265 . We should switch to released versions of dnet dependencies (after their release) and merge into master.

Issue #1251, opened 2021-03-19
Align on the new persistent identifier creation strategy introduced in the Oaf model


marekhorst:
More details on the strategy itself is available on redmine issue ticket:
https://issue.openaire.research-infrastructures.eu/issues/6486

przemyslawjacewicz:
Updates merged into temporary branch `przemyslawjacewicz_1251_1252_alignment_with_updates_in_oaf_model`. We should switch to released versions of dnet dependencies (after their release) and merge into master.

PR #1265, opened 2021-04-21
#1251 and #1252: Alignment with updates in OAF model

przemyslawjacewicz:
This PR aligns software and patent export with recent updates in OAF model. It does not close linked issues because new versions of `dhp-common` and `dhp-schemas` libs are not yet released and this PR uses snapshot versions created from `stable_ids` branch of `dnet-hadoop` project. The base branch of this PR is a temporary branch where the changes will wait for the release of dependencies. After the release we should make additional changes to use new versions of libs and then merge the changes into master branch of IIS.

First part of main changes is related to new strategy for values of `pid`, `instance.pid` and `instance.alternateIdentifiers` fields of exported entities. For patent entities we are clearing `pid` and `instance.pid` fields and setting `instance.alternateIdentifiers` field with pid values.  For software entities there are no changes because there are no pid values. 

Second part of main changes is related to identifier creation implementation for exported entities. We are switching to id creation based on `IdentifierFactory.createIdentifier` method in `dhp-common` lib. Previous ids are treated as fallback id. This change, however, does not really change the ids created upon export, because new strategy is based on pid values which are empty for both entities and in both cases `IdentifierFactory.createIdentifier` method uses fallback ids. This change mainly allows to use common strategy of id creation for entities. 

Patent and software exporting Spark jobs are refactored. The jobs are now quite similar in structure. Refactorization also allowed to find a difference in processing: software entities were exported regardless of the existence of matching document metadata. Now both jobs use the same processing: only entities that are related to existing metadata are exported. The processing is also simplified and cleaned up:
* software export was moved to dedicated package
* some unnecessary joins are removed
* variables names' simplification
* file moves and unification
* entity.properties -> entity.expectiations in tests


iis-wf/iis-wf-export-actionmanager/pom.xml
```diff
@@ -39,6 +39,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>eu.dnetlib.dhp</groupId>
+            <artifactId>dhp-common</artifactId>
+            <version>1.2.4-SNAPSHOT</version>
```
marekhorst, line 45: We need to make sure to remove this SNAPSHOT dependency and everything SNAPSHOT related.

marekhorst (approved):
LGTM. We just need to remember to revert all SNAPSHOT related changes into releases before merging with the master branch.

PR #1269, opened 2021-04-29
#1251and #1252: Alignment with updates in OAF model - dependencies update

przemyslawjacewicz:
This PR updates dnet dependencies after release of `dhp-schemas` lib. 

`dhp-common` dependency is removed because `IdentifierFactory` class was moved to `dhp-schemas`. I've used the same version range as in other dnet dependencies.


marekhorst (approved):
LGTM.

PR #1274, opened 2021-05-12
Closes #1272: Align InfoSpace importer with the recent updates in Oaf model related to the introduction of instance.alternateIdentifiers

### LINKED ISSUE #1272: Align InfoSpace importer with the recent updates in Oaf model related to the introduction of instance.alternateIdentifiers
This is a supplement of #1251. We need to introduce changes in the InfoSpace importer module as well.

Long story short: we should avoid relying on the `Entity#pid` field (to be deprecated soon) and read identifiers from `Result#instance#pid` and `Instance#alternateIdentifiers` fields. Those sets are disjoint due to the phase run before mining but to be on a safe side we should treat `Result#instance#pid` as the more trustable than `Instance#alternateIdentifiers` just to make sure we have a specific non-random rule of picking the ids.

The following importer modules are relying on pids right now:
* `DatasetMetadataConverter`
* `DocumentMetadataConverter`

The full correspondence in the pid context is available in this slack thread:
https://openairegroup.slack.com/archives/C8VNGBV5E/p1620227883048300

przemyslawjacewicz:
Note: OAF model changes implemented in this PR do not test well with our current data. Specifically avoiding the use of `Result#pid` field in converters and switching to `Result#instance#pid` or `Result#instance#alternateIdentifier` caused removal of some valid data from importer test outcome because our test data rely on `Result#pid` field. We could consider updating our test input to include changes in the model.

### DIFF
--- a/iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtils.java
+++ b/iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtils.java
@@ -1,17 +1,18 @@
 package eu.dnetlib.iis.wf.importer.infospace.converter;
 
-import java.time.Year;
-import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeParseException;
-import java.util.Collection;
-import java.util.List;
-import java.util.stream.Collectors;
-
+import eu.dnetlib.dhp.schema.oaf.Instance;
+import eu.dnetlib.dhp.schema.oaf.Result;
+import eu.dnetlib.dhp.schema.oaf.StructuredProperty;
+import eu.dnetlib.iis.wf.importer.infospace.approver.FieldApprover;
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.log4j.Logger;
 
-import eu.dnetlib.dhp.schema.oaf.StructuredProperty;
-import eu.dnetlib.iis.wf.importer.infospace.approver.FieldApprover;
+import java.time.Year;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * Static class with metadata conversion utilities.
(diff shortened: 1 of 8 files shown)


iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtils.java
```diff
@@ -60,5 +61,31 @@ private static Year parseYearOrGetNull(String year, Logger log) {
...
+
+        List<StructuredProperty> pid = result.getInstance().stream()
+                .map(x -> Optional.ofNullable(x.getPid()).map(Collection::stream))
+                .filter(Optional::isPresent)
+                .flatMap(Optional::get)
+                .collect(Collectors.toList());
+
+        if (CollectionUtils.isNotEmpty(pid)) {
+            return pid;
+        }
```
marekhorst, line 83: We should probably get as many identifiers as possible, from both sources (namely `instance#pid` and `instance#alternateIdentifier`). So we should not return results after inspecting the first source of ids only.

The logic I was describing in the main ticket ("treat `Result#instance#pid` as the more trustable than `Instance#alternateIdentifiers` just to make sure we have a specific non-random rule of picking the ids") was referring to dealing with duplicates whenever the same type of id occur in both places and we end up with an overlap. Sorry I didn't express that clearly enough.

When I was writing the #1272 description **I had the map type in mind** (where the duplicates resolution is a must) but since we are building a `List<StructuredProperty>` we are not forced to deal with duplicates, technically, so maybe we can just trust the disjoint sets and simply propagate the ids from both fields without the duplicates resolution.

marekhorst (changes requested):
Let's just make one change and include identifiers from both sources (more in comment).

Regarding the tests you have mentioned - I assume you are referring to integration tests where we evaluate the JSON output, right? I think if we have unit tests for the piece of code responsible for ids manipulation then we can pass the verification of pids in integration tests.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/converter/MetadataConverterUtils.java
przemyslawjacewicz: Ok, I got mislead by the slack conversation also. Fixed this, I've also added duplicates resolution, just to be on the safe side, and also because it was quite easy to do it with the a list.

przemyslawjacewicz:
Fixed pid resolution. 

> Regarding the tests you have mentioned - I assume you are referring to integration tests where we evaluate the JSON output, right? I think if we have unit tests for the piece of code responsible for ids manipulation then we can pass the verification of pids in integration tests.

Yes, that was my point - to make integration tests working I had to set some fields to `null` and I thought that the output JSONs start to look now as shells with empty data. But we've got unit tests for that so I also think we're ok.

marekhorst (approved):
LGTM.

PR #1276, opened 2021-05-13
Closes #1264: Align content identifiers with metadata record persistent identifiers

### LINKED ISSUE #1264: Align content identifiers with metadata record persistent identifiers
More details on the changes in Oaf model are available here:
https://issue.openaire.research-infrastructures.eu/issues/6510

przemyslawjacewicz:


iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJobUtils.java
```diff
@@ -0,0 +1,70 @@
...
+        org.apache.spark.sql.Dataset<Row> resultIdMapDF = sourceDatasetDS.select(col("id"), explode(col("originalId")).as("oid"))
+                .union(sourceOtherResearchProductDS.select(col("id"), explode(col("originalId")).as("oid")))
+                .union(sourcePublicationDS.select(col("id"), explode(col("originalId")).as("oid")))
+                .union(sourceSoftwareDS.select(col("id"), explode(col("originalId")).as("oid")))
+                .distinct();
+
+        org.apache.spark.sql.Dataset<Row> dedupRelationDF = new AvroDatasetSupport(spark).toDF(
+                spark.createDataset(dedupRelation.rdd(), Encoders.kryo(IdentifierMapping.class)), IdentifierMapping.SCHEMA$);
+
+        Dataset<Row> dedupRelationWithOidDF = dedupRelationDF
```
marekhorst, line 63: Just as we have already discussed today: we should change the joining order because we need to translate all the identifiers in `DocumentContentUrl` down the stream, not only deduped ones. So we need to have all entries from originalId->persistentId from the graph supplemented with persistend->dedupedId mapping wherever applicable (for each persistentId from the originalId->perisitentId map which could be found in persistend->dedupedId map).

More in:
https://issue.openaire.research-infrastructures.eu/issues/6510#note-24

marekhorst (changes requested):
Apart from the crucial change in joining order we should also rename all the occurrences of dedupmapping (into something more generic: identifierMapping or persistentIdentifierMapping or sth similar) occurring after the `mapToObjectStoreId()` method call which produces full identifiers map, not only the deduped ids. Not only in java code but also in workflow definitions etc.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJobUtils.java
przemyslawjacewicz: Yes, you're right I didn't get that at the beginning. But giving this more thought I think that just a change in join order is not enough because we can have three types of id mappings as the result:
* for deduplicated entities that are not `Result` types we should map using deduplication mapping only - this case is the same as what we currently do
* for `Result` types that are not deduplicated we should map using `Result` mapping only - in this case we should map using `id` and `originalId` fields of the `Result` entity
* for deduplicated entities of `Result` types  we should map using a combination of deduplication mapping and `Result` mapping - in this case we should map using `newId` field of deduplication mapping and `originalId` field of `Result` mapping

This means that we end up with a full outer join between both mappings with additional selection of proper 'ouput' fields based on the join result. Running this version of the code with recent PROD run results gives 95.347.794 total mapping records with 61.114.481 records matching the first case (only deduplication mapping), 18.736.545 records matching the second case (only `Result` mapping) and 15.496.768 records matching the last case (both mappings). I've updated the code to implement this change.

Secondly, I've added filtering to the `Result` mapping based on entity `id` field, because (strangely) I forgot to do that. I've also changed the name of the method because the old name was misleading.

przemyslawjacewicz:
> Apart from the crucial change in joining order we should also rename all the occurrences of dedupmapping (into something more generic: identifierMapping or persistentIdentifierMapping or sth similar) occurring after the `mapToObjectStoreId()` method call which produces full identifiers map, not only the deduped ids. Not only in java code but also in workflow definitions etc.

Done, I've used 'identifierMapping` name. Re-requesting review.

przemyslawjacewicz:
I've just pushed updates that were discussed offline which simplify object store id creation. 

Full outer join from deduplication mapping to `Result`-based mapping is replaced with left outer join with `Result`-based map on the left. 

I've also checked that `50|` and `dedup` filters cannot be removed - mapping counts with and without filters change. Here's a snippet from Zeppelin notebook
```scala
    val resultIdAndOriginalIdDF = datasetDF.select($"id", explode($"originalId").as("oid"))
      .union(otherResearchProductDF.select($"id", explode($"originalId").as("oid")))
      .union(publicationDF.select($"id", explode($"originalId").as("oid")))
      .union(softwareDF.select($"id", explode($"originalId").as("oid")))

    resultIdAndOriginalIdDF.count() //445744097

    resultIdAndOriginalIdDF.where(not($"id".like("%dedup%"))).count() //336059270

    resultIdAndOriginalIdDF.where($"oid".like("50|%")).count() //49630073

    resultIdAndOriginalIdDF.where(not($"id".like("%dedup%")) && $"oid".like("50|%")).count() //34219920
```
This might suggest that there are some issues with populating `originalId` field of `Result` types because as Claudio wrote in redmine ticket the `dedup` filter should be redundant.

Re-requesting code review.

przemyslawjacewicz:
I've just pushed test fixes.

marekhorst:
> This might suggest that there are some issues with populating `originalId` field of `Result` types because as Claudio wrote in redmine ticket the `dedup` filter should be redundant.

Does it mean the `originalId` sometimes does not start with `50|` prefix and it may include `dedup` phrase?

You are checking `50|` filter against `oid` field but `%dedup%` filter against different, `id` field. Was it intentional? 

I mean was your intention to verify whether `dedup` phrase is part of `originalId` or `persistenId`? I would definitely not expect the first one to be matched but would be ok for the second (assuming you are working on a full set of ids from the graph including new representatives).

claudioatzori:
> This might suggest that there are some issues with populating `originalId` field of `Result` types because as Claudio wrote in redmine ticket the `dedup` filter should be redundant.

When I mentioned that the filter on the `%dedup%` prefix should be redundant, I based my statement on the assumption that all the full-texts are associated with pre-dedup identifiers (in turn possibly mapped to dedup ids), therefore in my reasoning, not a single dedup id was involved, making them negligible. However, I do recognize I wrote that without knowing how the identifier resolution mechanism internally works in the IIS, so maybe instead of helping to clarify I created more confusion :/

Anyway, Indeed the population of the 'old' identifiers among the `result.originalId` filed was not complete. As I mentioned in [#6510#note-22](https://support.openaire.eu/issues/6510#note-22), all results from DOIBoost and Datacite did not include any of the old identifiers among the originalIds. Some days ago I took the initiative to modify the procedure to preserve those original identifiers in the doiboost construction process (https://code-repo.d4science.org/D-Net/dnet-hadoop/commit/1bd70fa2c620492cd4abb17d40128abc447f0d8d), so as for today, to my knowledge, among the results possibly associated with a full-text, only the records from Datacite does not include the old identifier among the originalIds.

przemyslawjacewicz:
> > This might suggest that there are some issues with populating `originalId` field of `Result` types because as Claudio wrote in redmine ticket the `dedup` filter should be redundant.
> 
> Does it mean the `originalId` sometimes does not start with `50|` prefix and it may include `dedup` phrase?

For the first part of the question - yes there are 396114024 (445744097−49630073) records in the mapping with `originalId` not starting with `50|` prefix. For the second part of the question - I don't know, because I didn't count the records. The filter with `dedup` phrase is applied on `id` field, just as the query referenced by Claudio in redmine ticket.

> You are checking `50|` filter against `oid` field but `%dedup%` filter against different, `id` field. Was it intentional?
> 
> I mean was your intention to verify whether `dedup` phrase is part of `originalId` or `persistenId`? I would definitely not expect the first one to be matched but would be ok for the second (assuming you are working on a full set of ids from the graph including new representatives).

Generally, the filters in current form are intentional because I've taken the filters from the query Claudio referenced in redmine ticket:
```sql
select distinct id, oid
from openaire_beta_20210414_stable_ids.result lateral view explode(originalid) o as oid
where oid like '50|%' and id not like '%dedup%';
```
The construction of this query (and our code build from it) matched the version of the mapping with 'full outer join'.

przemyslawjacewicz:
Giving the 'dedup' filter more though I think that the version of the mapping based on 'left outer join' shouldn't use it. The goal of the mapping is to have a mapping from the object store id to new, persistent id that definitely can a post-dedup id. The join should then only replace pre-dedup ids with post-dedup ids. @marekhorst What's your opinion ?

marekhorst:
> Giving the 'dedup' filter more though I think that the version of the mapping based on 'left outer join' shouldn't use it. The goal of the mapping is to have a mapping from the object store id to new, persistent id that definitely can be a post-dedup id. The join should then only replace pre-dedup ids with post-dedup ids.

Yes, this is true. 

The map you apply filtering on includes all the `Result` persistent identifiers and associated original identifiers. The reason I think dedup filtering is redundant is simply because the `persistentId`<->`orignialId` mapping is not propagated from deduplicated records to the new representatives (with `dedup` phrase in `persistentId`) therefore I would not expect to have any `dedup` entry in this map. Simple as that, but only if my assumption is valid. Well, this was the main reason we decided to dig deep in the graph to retrieve `persistentId`<->`orignialId` mappings from deduplicated records (with `deletedbyinference`=`true`) otherwise we could simply stick to the new representatives only and non-duplicate records already imported by IIS.

But if your analysis proves there are some `dedup` identifiers in `persistentId`<->`orignialId` map (as imported from the graph, before applying left join with dedupIds) then it means something unexpected is happening which might be worth digging into.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJobUtils.java
```diff
@@ -0,0 +1,80 @@
...
+     *
+     * @param dedupRelation              {@link IdentifierMapping} RDD containing mappings from deduplicated ids to original ids.
+     * @param sourceDataset              {@link eu.dnetlib.dhp.schema.oaf.Dataset} RDD.
+     * @param sourceOtherResearchProduct {@link eu.dnetlib.dhp.schema.oaf.OtherResearchProduct} RDD.
+     * @param sourcePublication          {@link eu.dnetlib.dhp.schema.oaf.Publication} RDD.
+     * @param sourceSoftware             {@link eu.dnetlib.dhp.schema.oaf.Software} RDD.
+     * @param spark                      Instance of SparkSession.
+     * @return An RDD of {@link IdentifierMapping} with object store ids.
+     */
+    public static JavaRDD<IdentifierMapping> produceObjectStoreId(JavaRDD<eu.dnetlib.dhp.schema.oaf.Dataset> sourceDataset,
```
marekhorst, line 39: I'd rename this method. Mostly because we struggle with a lot of confusion with this whole task and someone reading this code for the first time would be double-confused. Something like `produceGraphIdToObjectStoreIdMapping` or similar would be OK, just to indicate what it really does. I know it is already in method description but it will be easier to understand the code referencing this method.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJobUtils.java
```diff
@@ -0,0 +1,80 @@
...
+
+        Column idIsNotDedup = not(col("id").like("%dedup%"));
+        Column oidIsResultType = col("oid").like(InfoSpaceConstants.ROW_PREFIX_RESULT + "%");
+
+        Dataset<Row> resultIdMapDF = sourceDatasetDS.select(col("id"), explode(col("originalId")).as("oid"))
+                .union(sourceOtherResearchProductDS.select(col("id"), explode(col("originalId")).as("oid")))
+                .union(sourcePublicationDS.select(col("id"), explode(col("originalId")).as("oid")))
+                .union(sourceSoftwareDS.select(col("id"), explode(col("originalId")).as("oid")))
+                .where(idIsNotDedup)
+                .where(oidIsResultType)
```
marekhorst, line 62: Let's discuss the necessity of this filtering (already mentioned in PR comments).

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJobUtils.java
```diff
@@ -0,0 +1,80 @@
...
+                .distinct();
+
+        Dataset<Row> dedupRelationDF = new AvroDatasetSupport(spark).toDF(
+                spark.createDataset(dedupRelation.rdd(), Encoders.kryo(IdentifierMapping.class)), IdentifierMapping.SCHEMA$);
+
+        Column dedupMappingNotMatched = col("newId").isNull().and(col("id").isNotNull());
```
marekhorst, line 68: Do we need to check if `id` is not null if we left join on `id`? It will be always there.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJobUtils.java
```diff
@@ -0,0 +1,80 @@
...
+
+        Dataset<Row> identifierMappingDF = resultIdMapDF
+                .join(dedupRelationDF, col("id").equalTo(col("originalId")), "left_outer")
+                .select(
+                        when(dedupMappingNotMatched, col("id"))
+                                .otherwise(col("newId")).as("newId"),
+                        col("oid").as("originalId")
+                );
```
marekhorst, line 76: It is quite difficult to verify the (rather simple) logic behind this piece of code, possibly because of the brackets I need to count in order to know on which object each method is called. Maybe new lines or using temp variables could help a little bit to increase the readability?

One another thing which could improve the readability: why using condition with inclusive negation here? Usually it is much easier to rely on `dedupMappingMatched` condition instead of `dedupMappingNotMatched` especially when you have `otherwise` as a next operator in a sequence.

iis-wf/iis-wf-import/src/test/resources/eu/dnetlib/iis/wf/importer/infospace/output/report.json
```diff
@@ -2,6 +2,6 @@
 {"key":"import.infoSpace.dataset", "type":"COUNTER", "value": "2"}
 {"key":"import.infoSpace.docProjectReference", "type":"COUNTER", "value": "1"}
 {"key":"import.infoSpace.project", "type":"COUNTER", "value": "1"}
-{"key":"import.infoSpace.docDedupDocReference", "type":"COUNTER", "value": "1"}
+{"key":"import.infoSpace.docIdentifierMappingDocReference", "type":"COUNTER", "value": "0"}
```
marekhorst, line 5: I know this is an easy way to make test working again ;) 

But in order to be sure we are still able to import identifier mapping we should update the test input to make sure identifier mapping importer works as it is supposed to work.

marekhorst (changes requested):
Some changes are straightforward as requested in the comments but some of them would need to be discussed after the daily meeting.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJobUtils.java
przemyslawjacewicz: We do not need to, it's a residue of the full outer join.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJobUtils.java
przemyslawjacewicz: There are only two brackets grouped at max, so it's not that much ;-) `when`/`otherwise` is like simple 'if' here so there's not much place for increasing readability or using temp variables. I've put the whole statement on one line, I think this helps a bit.

przemyslawjacewicz:
One thing we need to figure out is how to deal with deduplicated records and do we need a filter or some other processing to deal with them.

I've checked what's inside `originalId` field of deduplicated records:
```shell
    val resultIdAndOriginalIdDF = datasetDF.select($"id", explode($"originalId").as("oid"))
      .union(otherResearchProductDF.select($"id", explode($"originalId").as("oid")))
      .union(publicationDF.select($"id", explode($"originalId").as("oid")))
      .union(softwareDF.select($"id", explode($"originalId").as("oid")))
      .distinct()
      .cache()

    resultIdAndOriginalIdDF.count() //445744094

    println(
      resultIdAndOriginalIdDF.where($"id".like("%dedup%")).toJSON.take(20). mkString("\n")
    )
/*
{"id":"50|doi_dedup___::0c834c67556d7926305f017018d08fe1","oid":"10.5281/zenodo.2418082"}
{"id":"50|doi_dedup___::4461b4db6bc7dea7b9209974962aa7e2","oid":"oai:figshare.com:article/4089471"}
{"id":"50|doi_dedup___::d8972f1b1c06bf018d9641e86f910ec5","oid":"10.5061/dryad.n8f30/2"}
{"id":"50|doi_dedup___::05c87aaf566fb65045e74ae2452836cf","oid":"50|datacite____::0de8d87366988315d52cae380d310683"}
{"id":"50|doi_dedup___::3fe731448c06595fb57c339ff13f9e32","oid":"4699a1e5-3e8d-485b-902a-4c1480d3e39e"}
{"id":"50|doi_dedup___::7a81ebee2541467c121fa4331dac81b6","oid":"50|datacite____::7a81ebee2541467c121fa4331dac81b6"}
{"id":"50|doi_dedup___::c6f60d5bcc54d452f92edb12e154097f","oid":"10.6084/m9.figshare.5650198"}
{"id":"50|doi_dedup___::d24abfe77bc89fd4fcceb77a47bd376b","oid":"10.6084/m9.figshare.13642081"}
{"id":"50|doi_dedup___::041bd4c054a1e93d14db548bc791d079","oid":"50|datacite____::b8304d7f1d1abbe56908be51dd449a1b"}
{"id":"50|doi_dedup___::8dd5dd943e3aa22805ee9d0c8df12c5a","oid":"10.5281/zenodo.1479442"}
{"id":"50|doi_dedup___::02c8dbc0953b89c475ac56c0350770bb","oid":"10.3932/ethz-a-000599146"}
{"id":"50|doi_dedup___::37c0c055d3553141ef0977de67ecf8be","oid":"10.21233/n3jp99"}
{"id":"50|doi_dedup___::7918a41648f6379a5af9a6ac09f1923d","oid":"50|datacite____::d4193b2c31d4c79f0cbc58c36176bc9b"}
{"id":"50|doi_dedup___::7da6fffbd1b6bbc066f657577e8cfe98","oid":"50|datacite____::9ed0ee7324ad7070c6f8b03011713ed8"}
{"id":"50|dedup_wf_001::963e742bae97e71046e7155bd6b6c4e0","oid":"oai:figshare.com:article/2447926"}
{"id":"50|doi_dedup___::65e8c665df438cd8af1d8ce22d8a1f70","oid":"oai:pangaea.de:doi:10.1594/PANGAEA.868690"}
{"id":"50|dedup_wf_001::001946d2cb97e91e866b28d67ae2a995","oid":"PRJNA309753"}
{"id":"50|doi_dedup___::03fe3ec1dc783ad5538b5d7910d76269","oid":"410012005-004"}
{"id":"50|doi_dedup___::519ccf75f459842d0a129b4c8a8de8e0","oid":"oai:figshare.com:article/1122783"}
{"id":"50|doi_dedup___::7bf150b3c145564ad7b6e23d7d5e0580","oid":"10.6084/m9.figshare.57455"}
*/
```
There are various identifiers (like DOI) that are not what we're looking for because they cannot be the object store ids (and should be filtered out using `50|` filter) and there are openaire ids that are interesting for us. To check what's inside those ids, that is if they are the object store ids propagated from pre-deduplication records (or not), I've joined the deduplication records with deduplication mapping to get the pre-deduplication id and then self-joined with the same mapping to get the `oid` field again. The goal was to check how `oid` field of deduplicated records relate to `oid` field of pre-deduplicaton records. Here are the results:
```shell
    val relationDS = relationDF.as[Relation](Encoders.bean(classOf[Relation]))

    /**
     * Verifies whether given relation should be accepted according to its characteristics.
     */
    def acceptRelation(relation: Relation): Boolean =
      "resultResult" == relation.getRelType && "dedup" == relation.getSubRelType && "merges" == relation.getRelClass

    val approver = new DataInfoBasedApprover("iis", true, null)
    val converter = new DeduplicationMappingConverter

    val dedupMappingDS = relationDS
      .filter(x => acceptRelation(x))
      .filter(x => approver.approve(x))
      .map(x => converter.convert(x))(Encoders.kryo(classOf[IdentifierMapping]))

    val dedupMappingDF = new AvroDatasetSupport(spark).toDF(dedupMappingDS, IdentifierMapping.SCHEMA$)

    val zz = resultIdAndOriginalIdDF.where($"id".like("%dedup%") && $"oid".like("50|%")).as("left")
      .join(dedupMappingDF, $"id" === $"newId")
      .join(resultIdAndOriginalIdDF.where($"oid".like("50|%")).as("right"), $"originalId" === $"right.id")
      .select($"left.id", $"left.oid", $"right.oid")
      .distinct()

    zz.count() //69440650

    zz.where($"left.oid" =!= $"right.oid").distinct().count() //54030503
```
So it seems that there are deduplicated records with `oid` field different then `oid` field of pre-deduplication record. This would suggest that we have to have different processing for deduplicated records (to get the `oid` of pre-deduplication record) and for not deduplicated records (to get the `oid` field of the record). @claudioatzori  @marekhorst  What do you think ?

marekhorst:
> I've checked what's inside originalId field of deduplicated records:

I take it as a mistake (or just a mental shortcut), but you are not checking what is inside deduplicated records but their representatives. Just to make our vocab straight :)

> So it seems that there are deduplicated records with `oid` field different then `oid` field of pre-deduplication record. 

OK, maybe there is a reason for that: like some of them are not propagated or the new representative is supplemented with new ids not coming from deduped records. I guess there is much more happening than we need to know for our IIS mapping purpose so I would leave it for @claudioatzori to pick up if he finds anything suspicious.

Your example clearly shows `originalId` may convey various formats of identifiers and it may be set for new representatives as well (with `dedup` phrase in id). So this proves we need to keep both filters (`50|` on `originalId` and `dedup` on `id`) in our query. 

What I also wanted to get from the data inspection was to confirm that we really need to take `originalId` from deduplicated records not from the new representative record (because allegedly they are not propagated there but I could find direct confirmation in [#6510](https://issue.openaire.research-infrastructures.eu/issues/6510)). Were you able to confirm that? 

Let's not throw ourselves into full analysis of `originalId`s because we would never end this task. There is always something behind the corner...

> This would suggest that we have to have different processing for deduplicated records (to get the `oid` of pre-deduplication record probably) and for not deduplicated records (to get the `oid` field of the record).

@przemyslawjacewicz, isn't it what we already do? 

1. we get the whole bag of `Result` entities (including deduped ones and the new representatives and of course the ones which do not have duplicates)
2. we build the id<->oid map
3. we filter out all entries where either `id` is `dedup` (because it does not bring anything to the table) or `oid` is not an OpenAIRE id (because we don't care about other types). So we have OpenAIRE ids on both ends, no `dedup` id on either side.
4. we join this map with dedupmapping to translate OpenAIRE ids of deduplicated records into the new representatives (because deduplicated records ids are not present in the IIS datastores)

and we're done.

I am pretty confident the mapping is ok, so let's push it forward, merge and deploy the new IIS version.

iis-wf/iis-wf-import/src/main/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJobUtils.java
marekhorst: OK, we need both of them.

przemyslawjacewicz:
Following offline discussion regarding propagation of `originalId` to new representatives I've done one more (maybe final) check. Just as in the previous comment I've compared `originalId` of new representatives with `originalId` of original records but this time taking duplicates into account. Here are the results:
```scala
    val resultIdAndOriginalIdDF = datasetDF.select($"id", explode($"originalId").as("oid"))
      .union(otherResearchProductDF.select($"id", explode($"originalId").as("oid")))
      .union(publicationDF.select($"id", explode($"originalId").as("oid")))
      .union(softwareDF.select($"id", explode($"originalId").as("oid")))
      .distinct()

    val z = resultIdAndOriginalIdDF.where($"id".like("%dedup%") && $"oid".like("50|%")).as("left")
      .join(dedupMappingDF, $"id" === $"newId")
      .join(resultIdAndOriginalIdDF.where($"oid".like("50|%")).as("right"), $"originalId" === $"right.id")
      .select($"left.id".as("id"), $"left.oid".as("dedup_oid"), $"right.oid".as("oid"))
      .distinct()

    z
      .select($"id")
      .distinct()
      .count() //10612530

    z
      .groupBy($"id")
      .agg(collect_set($"dedup_oid").as("dedup_oid_list"), collect_set($"oid").as("oid_list"))
      .where(array_sort($"dedup_oid_list") =!= array_sort($"oid_list"))
      .count() //12047
```

So there are 12047 new representatives with `originalId` content different then the `originalId` content of original records. Manual sampling of the records show that all `originalId` ids of new representatives are among the ids of original records:
```scala
    println(
      z
        .groupBy($"id")
        .agg(collect_set($"dedup_oid").as("dedup_oid_list"), collect_set($"oid").as("oid_list"))
        .where(array_sort($"dedup_oid_list") =!= array_sort($"oid_list"))
        .toJSON
        .take(5)
        .mkString("\n")
    )
/*
{"id":"50|doi_dedup___::02b4814e23f6462b03e447d13734a667","dedup_oid_list":["50|datacite____::3e1ee88268f2a17913da1d53af9563ab","50|datacite____::a50029f2d21ec300fb15a5df509095b4"],"oid_list":["50|datacite____::75800d41a9925a17bc73f32a54c7fe08","50|datacite____::daeea79729573d5b65eff0d2c3212b21","50|datacite____::a34210f275b55bbd8c257e82d751e827","50|datacite____::3e1ee88268f2a17913da1d53af9563ab","50|datacite____::a50029f2d21ec300fb15a5df509095b4"]}
{"id":"50|doi_dedup___::0460e1e49ee0bd3f2d59a379322c3980","dedup_oid_list":["50|od_______267::f3a5538a0b9be94891c48c73f3790352"],"oid_list":["50|datacite____::0f7fd83446da3e4eae1952eb56606885","50|od_______267::f3a5538a0b9be94891c48c73f3790352"]}
{"id":"50|doi_dedup___::049aae2470088885c6d1807bfb248bc0","dedup_oid_list":["50|datacite____::049aae2470088885c6d1807bfb248bc0"],"oid_list":["50|datacite____::63564f2367fd569e21d01ecb92072852","50|datacite____::f25c4325212a2e1a534895f7dcd5c2a7","50|datacite____::049aae2470088885c6d1807bfb248bc0"]}
{"id":"50|doi_dedup___::05b30dbbd661831b408eba8b712b72fe","dedup_oid_list":["50|datacite____::5a3aed358a13df0241a99444091b78e6","50|datacite____::05b30dbbd661831b408eba8b712b72fe"],"oid_list":["50|datacite____::5a3aed358a13df0241a99444091b78e6","50|datacite____::506c2e5350fab33045bae18adcb0a963","50|datacite____::8f89844f05cb498079fef7e82504744f","50|datacite____::05b30dbbd661831b408eba8b712b72fe"]}
{"id":"50|doi_dedup___::06fe95d1ee9da0c85eb5b2ca3f404bec","dedup_oid_list":["50|datacite____::211027bce114458d10f1ade3d3fa8821"],"oid_list":["50|datacite____::211027bce114458d10f1ade3d3fa8821","50|datacite____::0aa4e1a83c52f66cacb3586f1524c246"]}
*/
```
The number of records that differ is low - 12047 out of 10612530. So, as discussed offline, at this point we should move along with current identifier mapping and check IIS results after provisioning round.

przemyslawjacewicz:
Changes applies, re-requesting code review.

marekhorst (approved):
Looks good, let's do the release! :)

PR #1283, opened 2021-05-20
Closes #1280: Remove compression for avro datastores saved using spark-avro lib

### LINKED ISSUE #1280: Remove compression for avro datastores saved using spark-avro lib
Avro datastores saved using `AvroDataFrameWriter` and `AvroDatasetWriter` use default snappy compression. This integrates well with `SparkAvroLoader` and `SparkAvroSaver` from spark-utils but does not integrate with cmd line tools i.e.
* hdfs dfs -text
* avro2json
* run_hive.sh

We should remove compression when saving. This should be done by adding this option when saving 
```scala
option("compression", "uncompressed")
```
Integration with cmd line tools should be tested afterwards.

przemyslawjacewicz:
This PR removes compression when saving dataframes of Avro types. Avro dataset writer is also slightly improved, to reduce code duplication.

NOTE: there are no unit tests for checking the compression of saved Avro dataframes. This is because the compression does not affect the support of Avro datastores in IIS - we can read and write Avro datastores with and without compression without any issues using either spark-utils or spark-avro libs. The problem with compression manifests itself when external tools are used to process Avro files, like `hdfs dfs -text` or `avro2json`. The changes were manually tested using those tools.

### DIFF
--- a/iis-common/src/main/scala/eu/dnetlib/iis/common/spark/avro/AvroDataFrameWriter.scala
+++ b/iis-common/src/main/scala/eu/dnetlib/iis/common/spark/avro/AvroDataFrameWriter.scala
@@ -32,6 +32,7 @@ class AvroDataFrameWriter(df: DataFrame) extends Serializable {
       .write
       .format("avro")
       .option("avroSchema", avroSchema.toString)
+      .option("compression", "uncompressed")
       .save(path)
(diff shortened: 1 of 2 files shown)


marekhorst (approved):
LGTM.

PR #1314, opened 2022-02-17
Add Jenkinsfile.dockerized for running on the new Jenkins instance

mpol:
Based on the current Jenkinsfile but uses docker agents with a Maven
3/JDK 8 image.

Mount host /etc/passwd because the image is run by Jenkins as an UID
that is not present in the passwd file inside, and this actually
causes a few tests to fail (because of getpwuid being used by JVM in
some place). Also set user.home to the workspace directory (the one in
host passwd is not suitable).

Add pollSCM trigger until github-webhook is handled by the new
Jenkins.

Use direct credentials support for the sonar token and hardcode the sonar
URL (actually the same as the sonar name hardcoded in the original
Jenkinsfile), because the sonar plugin is unavailable.

Split build into Build and Sonar stages (the latter only for the
master branch) instead of handling the condition in the shell script.
Gives a nicer display but requires adding -DskipTests -DskipITs to the
second maven run.

In the build step do install, not package - apparently the original
Jenkinsfile actually depends on the installation being the sideeffect
of another build (without install iis-wf-primary tries to download
eu.dnetlib.iis:iis-wf-report:jar:tests artifact from artifactory, but
nothing is actually published there).

Replace the unavailable warnings step with recordIssues from the
warnings-ng plugin.


marekhorst (approved):
LGTM.

PR #1320, opened 2022-03-11
Change complete_taxonomies.py syntax to Python 3

mpol:
In Python 3 print is a function not a statement and requires parentheses.


marekhorst (approved):
OK.

PR #1325, opened 2022-03-17
Fix the final categorization-regex for "NRC".

LSmyrnaios:


marekhorst:
Rebased, squashed, supplemented with project reference extraction NRC test case and merged with IIS master branch with this commit: https://github.com/openaire/iis/commit/3abf76c5ae2f1c57f5975218cce5956d842c4421.

PR #1328, opened 2022-03-23
Use https for maven.ceon.pl

mpol:
Change http URLs in poms to https.
http access is currently broken on maven.ceon.pl and there is no
reason not to use https here.


marekhorst (approved):
LGTM.

PR #1338, opened 2022-04-12
Closes #1336: Rename citatations to references in ExtraInfo element

### LINKED ISSUE #1336: Rename citatations to references in ExtraInfo element
Originally requested on redmine: https://support.openaire.eu/issues/7509

`CitationsActionBuilderModuleFactory` should be reimplemented and current `citatations` and `citatation` wording should be changed to `references` and `reference` respectively.

marekhorst:
This is a simple change: renaming `citations` element into `references` and `citation` into `reference` as specified in redmine: 

https://support.openaire.eu/issues/7509?next_issue_id=7507&prev_issue_id=7510#note-10


iis-common/src/main/java/eu/dnetlib/iis/common/model/extrainfo/ExtraInfoConstants.java
```diff
@@ -11,9 +11,9 @@ public final class ExtraInfoConstants {
 	public static final String NAME_RESULT_STATISTICS = "result statistics";
 	public static final String NAME_AUTHOR_STATISTICS = "author statistics";
 	
-	public static final String NAME_CITATIONS = "result citations";
+	public static final String NAME_REFERENCES = "result references";
 	
-	public static final String TYPOLOGY_CITATIONS = "citations";
+	public static final String TYPOLOGY_REFERENCES = "references";
```
mpol, line 16: I get that this is what was specified (as much as anything was) in the original redmine ticket. But I wonder - having no actual idea - is this typology certain not to clash with something else that we also call references? Maybe it should be bibliographic-references or something similar to be more specific?

mpol (approved):
LGTM as specified. I ask one question about the specification itself in a code comment.
I did not spend too much time looking for any missing parts but the test should make sure there are none.

I would suggest noting/explaining the lack of change in class names etc. in the commit message.

iis-common/src/main/java/eu/dnetlib/iis/common/model/extrainfo/ExtraInfoConstants.java
marekhorst: I agree, this i a good point, I have created a note on that matter on redmine: https://support.openaire.eu/issues/7509#note-16.

marekhorst:
Just pushed another commit aligning the `ExtraInfo` payload name and typology with our agreement in https://support.openaire.eu/issues/7509#note-20.

mpol (approved):
LGTM

marekhorst:
Rebased, squashed and merged with this commit: https://github.com/openaire/iis/commit/7b7ad045bdde8597a8f3f9e509556625b6af50c0.

Issue #1339, opened 2022-04-13
Perform test mining for Ariadne resources


marekhorst:
Originally requested on redmine: https://support.openaire.eu/issues/7685

We should start with a dedicated branch where we could place Ariadne mining code and run in isolation. Without integrating it with IIS main workflow and exporting an outcome to ActionManager.

marekhorst:
The project ended, so did testing activities.

PR #1340, opened 2022-04-13
upload ariadne mining script

johnfouf:


marekhorst:
As already indicated in #1339 The project ended, so did testing activities.

PR #1344, opened 2022-04-21
Closes #1326: Documents similarity fails when run on the non-deduplicated graph

### LINKED ISSUE #1326: Documents similarity fails when run on the non-deduplicated graph
Originally reported on redmine: https://issue.openaire.research-infrastructures.eu/issues/7333#note-12

Here is the failure example:

http://iis-cdh5-test-m3.ocean.icm.edu.pl:19888/jobhistory/tasks/job_1614993033499_83511/m

all failed attempts had the following entry in the log:

```
Timed out after 600 secs
```

what indicates timeout issue.

The script failed at `coansys-integrated-similarity_docsim-recalculate-sim-about-1-filter` docsim subworfklow during the `sim1_postprocess_s1_e1_filter_input` phase (`sim1-postprocess-s1-e1-filter-sims.pig` script).

We might need to increase the timeout by setting `mapreduce.task.timeout` to a higher value than the default 10 minutes. The problem is docsim sources are maintained in coansys project so we need to hotpatch the sources (what we already do to cope with slightly [different docsim issue](https://support.openaire.eu/issues/2976).

We should run several experiments to make sure we need to increase the timeout just in this single place and to find the proper value (we could start with 30 mins).

marekhorst:
Increasing xmx memory assigned to documents similarity PIG jobs from `10GB` up to `12GB`.

### DIFF
--- a/iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/processing/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/processing/oozie_app/workflow.xml
@@ -137,7 +137,7 @@
         </property>
         <property>
             <name>ds_mapredChildJavaOpts</name>
-            <value>-Xmx10g</value>
+            <value>-Xmx12g</value>
             <description>mapred child java opts</description>
         </property>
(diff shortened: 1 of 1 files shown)


mpol (approved):
OK

PR #1347, opened 2022-05-04
Allow specific or all Canadian funders to be excluded from mininng.

LSmyrnaios:
- Avoid collecting specific (or all) canadian results, in case the operator specifies it (This helps with testing other funders quicker).
- Optimize and cleanup the "multiple_canadian_funders.py".
- Avoid having, by default, the Canadian unspecified-grantIDs inside the "grants" table of "base_projects.db".


marekhorst:
Merged with this rebased commit: https://github.com/openaire/iis/commit/30d34dd1eb5dcdb863fda02dda792f339534580a.

Issue #1329, opened 2022-03-25
Affiliation matching unit tests last far too long


marekhorst:
Apparently one of the long lasting tests, `AffOrgMatchVoterStrengthEstimatorAndTest`, was not marked with `@SlowTest` which makes this test to be run along with other unit tests instead of as a part of integration test suite which is executed nightly.

We should mark this test with `@SlowTest`.

marekhorst:
`AffOrgMatchVoterStrengthEstimatorAndTest` exends `TestWithSharedSparkContext` which is already marked with `@SlowTest` so we should check why this test is still being run as a unit test.

marekhorst:
It is directly related to already reported issue: #1193.

marekhorst:
An issue related to introducing `@SlowTests`: https://github.com/openaire/iis/issues/1166

we should inspect why it is not being excluded from execution during the `test` phase.

mpol:
So it seems the plan in #1166 was to exclude slow tests from standard builds.
Finally it was not done by default, but maybe we should use `-DotherExcludedGroups=SlowTest` in the IIS Jenkinsfile?

marekhorst:
> So it seems the plan in #1166 was to exclude slow tests from standard builds.
> Finally it was not done by default, but maybe we should use `-DotherExcludedGroups=SlowTest` in the IIS Jenkinsfile?

Yes, this is the way to go.

PR #1352, opened 2022-05-11
Do not run SlowTest tests during regular Jenkins builds

mpol:
As discussed in issue #1329 we do not want to run tests marked as
slow, some of which take dozens of minutes to finish, during regular
IIS builds. It is enough that they are run together with integration
tests and during deployment builds.

So add -DotherExcludedGroups=SlowTest to the maven invocation.

Additional not strictly related small fix - skipTests implies skipITs
so do not define both in the maven invocation for sonar.

### DIFF
--- a/Jenkinsfile.dockerized
+++ b/Jenkinsfile.dockerized
@@ -28,7 +28,7 @@ pipeline {
                 }
             }
             steps {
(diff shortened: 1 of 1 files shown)


marekhorst (approved):
LGTM.

PR #1355, opened 2022-05-20
Closes #1354:  DocumentTextCacheStorageUtilsTest breaks other tests if run at the wrong moment

### LINKED ISSUE #1354: DocumentTextCacheStorageUtilsTest breaks other tests if run at the wrong moment
Another case of tests interacting unexpectedly.

`DocumentTextCacheStorageUtilsTest` mocks Hadoop `Configuration` and when during one test
`DocumentTextCacheStorageUtils#storeInCache` throws an exception a `FileSystem` instance using
this mock is created and cached. Then later tests using `Filesystem#get` will get this mock-configured
filesystem from cache and likely fail - usually because it breaks checksumming configuration
by setting bytes per checksum to 0 (mocks return 0 from methods by default,
if we used Hadoop version with https://issues.apache.org/jira/browse/HADOOP-12831 fixed then I think
the test in `DocumentTextCacheStorageUtilsTest` would fail instead).

marekhorst:
Replacing Configuration mock with the new Configuration object instance.


mpol (approved):
LGTM. All the tests now pass.

PR #1357, opened 2022-05-23
Deploy IIS snapshots to our Maven repository

mpol:
Since other projects (in particular iis-examples) depend on some IIS
snapshot artifacts we need to upload them to the repository to allow for
independent builds.

Use the same approach as with MDE:
- Only upload from master branch, doing it in the same step that runs
the Sonar analysis.
- Pass the upload credentials using a maven settings file.


marekhorst (approved):
LGTM.

PR #1359, opened 2022-05-23
Run tests in a fully pseudo-random order

mpol:
We discovered several cases of unexpected test interdependencies
recently by tests failing in new environments.
Try running tests in a fully pseudo-random oder to detect other such
cases earlier.

Use test class and method ordering options provided by JUnit 5.
Bump JUnit version to the latest one since the one we used before did
not support test class reordering.

### DIFF
--- a/Jenkinsfile.dockerized
+++ b/Jenkinsfile.dockerized
@@ -28,7 +28,12 @@ pipeline {
                 }
             }
             steps {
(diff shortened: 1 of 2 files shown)


mpol:
We decided to do this in https://github.com/openaire/iis/issues/1354#issuecomment-1132872249

Issue #1298, opened 2021-07-30
Refactor the way IIS imports contents by dumping ObjectStore in favour of newly introduced PDF aggregation service


marekhorst:
Originally requested on redmine: 

https://support.openaire.eu/issues/6884

Issue #1342, opened 2022-04-20
Fix content_url_chain subworkflow actions order for cache builder workflow


marekhorst:
#1106 brought a bug which shows up only when running `content_url_chain` as a part of `metadataextraction_cache_builder` workflow which does not rely neither on deduplication mapping nor id filtering, both datastores are derived from the OpenAIRE Graph dump which is not available when rebuilding the metadataextraction cache.

The cumbersome way the execution path is controlled led to invalid action being excuted when id filtering should not be performed: `content-url-dispatcher` was set, `content-url-dedup` should be defined.

Issue #1360, opened 2022-06-29
Enable and refine the aggregation subsystem based importer integration test after installing Hive on CI test cluster


marekhorst:
After integrating the https://github.com/openaire/iis/blob/marekhorst_1298_refactor_iis_content_importer branch with #XXX pull request we end up with one  `eu.dnetlib.iis.wf.importer.content.WorkflowTest#testImportContentUrlWorkflowWithHiveBasedAggregationSubsystemAsBackend()` test as disabled because it can't be run on CI test cluster due to missing Hive service. It works on the main IIS CDH5 cluster though.

Apart from enabling the test once the hive service is installed on CI test cluster one another issue should be addressed related to the [`workflow.xml`](https://github.com/openaire/iis/blob/marekhorst_1298_refactor_iis_content_importer/iis-wf/iis-wf-import/src/test/resources/eu/dnetlib/iis/wf/importer/content_url/chain/agg_subsystem_based_importer/oozie_app/workflow.xml) definition (link currently pointing to the branch but it will be merged with IIS master branch eventually) which should be modified, along with the build system configuration in `iis-wf/pom.xml` file, to avoid setting `jdbc-url` and `hive_metastore_uris` to explicit values bound to a particular environment. 

The oozie workflow, by design, should remain environment-agnostic and it should be possible to rely on specific placeholders predefined for a given environment in `~/.iis/integration-test.properties` file stored in an environment just as it is already done for various env specific properties such as `sparkExecutorMemory` or `oozieServiceLoc`.

Currently defined values which are currently pointing to the main hadoop cluster and which should be addressed:
* `jdbc-url`: `jdbc:hive2://iis-cdh5-test-m3.ocean.icm.edu.pl:10000`
* `hive_metastore_uris`: `thrift://iis-cdh5-test-m3.ocean.icm.edu.pl:9083`

PR #1361, opened 2022-06-29
Integrate new aggregation subsystem based content importer module

marekhorst:
This pull request closes https://github.com/openaire/iis/issues/1298 by introducing support for importing `DocumentContentUrl` records from the newly introduced hive-based content aggregation subsystem.

The aggregation subsystem support is automatically enabled after providing explicit `import_content_pdfaggregation_table_name` parameter value at runtime. When the parameter is unspecified then content importer module works in legacy mode (objectstore compatible).

This pull request includes `agg_subsystem_based_importer` integration test which works properly on the main ocean cluster but cannot be run on a dedicated CI cluster due to the missing hive-service therefore it is currently disabled in `eu.dnetlib.iis.wf.importer.content.WorkflowTest` class and should be reenabled after installing Hive service on CI cluster (dedicated issue covering this task was already crated: #1360).


iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/content/WorkflowTest.java
```diff
@@ -13,10 +13,18 @@
...
-    public void testImportContentUrlWorkflow() {
+    public void testImportContentUrlWorkflowWithObjectStoreAsBackend() {
         OozieWorkflowTestConfiguration wfConf = new OozieWorkflowTestConfiguration();
         wfConf.setTimeoutInSeconds(720);
-        testWorkflow("eu/dnetlib/iis/wf/importer/content_url/chain/sampletest", wfConf);
+        testWorkflow("eu/dnetlib/iis/wf/importer/content_url/chain/objectstore_based_importer", wfConf);
+    }
+    
+    //@Test
+    //TODO reenable this test once Hive is installed on CI Test hadoop cluster
```
mpol, line 23: A perfect case for [`@Disabled("TODO reenable this test once Hive is installed on CI Test hadoop cluster")`](https://junit.org/junit5/docs/current/user-guide/#writing-tests-disabling), I think.

Even [`@DisabledIf`](https://junit.org/junit5/docs/current/user-guide/#writing-tests-conditional-execution-custom) could be fun if Hive is easily detectable, but that's not necessarily a good idea.

iis-wf/iis-wf-import/src/main/resources/eu/dnetlib/iis/wf/importer/content_url/chain/oozie_app/workflow.xml
```diff
@@ -146,9 +161,16 @@
 	        <arg>org.apache.avro.Schema.Type.STRING</arg>
 	        <capture-output />
 	    </java>
-	    <ok to="objectstore-content-url-importer" />
```
marekhorst, line 149: This particular change could be confusing but it is bound to a [specific commit](https://github.com/openaire/iis/pull/1361/commits/2439e414f8b591fc81df65a952b2813c726fa5c8) which is closing the #1342 issue spotted when working on #1298. 

Maybe it is not worth to elevate it to a dedicated PR but we should find a nice way to indicate this solution somehow in the commit comment once the commits get squshed before the merge.

mpol (approved):
LGTM. I think I understand what's going on here.

iis-wf/iis-wf-import/src/main/resources/eu/dnetlib/iis/wf/importer/content_url/chain/oozie_app/workflow.xml
```diff
@@ -239,7 +280,7 @@
 	
 	<decision name="decision-existence-filter">
         <switch>
-            <case to="content-url-dispatcher">${input_id eq "$UNDEFINED$"}</case>
```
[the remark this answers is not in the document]
mpol: Yes, I was surprised at first but found the reason for this change.

I personally prefer just not squashing, but adding "Also closes #1342, which was found during this work, by going to content-url-dedup from decision-existence-filter in the content_url_chain" should be enough, I think.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/content/HiveBasedDocumentContentUrlImporterJobTest.java
```diff
@@ -0,0 +1,128 @@
...
+    private Path outputReportDir;
+    
+    private HiveBasedDocumentContentUrlImporterJobTest() {
+        // enabling hive support in spark session instantiated for this set of tests
+        super(true);
+    }
+    
+    @BeforeEach
+    public void beforeEach() {
+        super.beforeEach();
```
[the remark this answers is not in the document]
marekhorst: You're right. I didn't pay attention to this, just copy&pasted from some other test class. 

I guess it might be some leftover when the IIS test code base was switched from junit4 to junit5 as it seems the default behavior [has changed](https://stackoverflow.com/questions/48864716/before-beforeeach-inheritance-behaviour-change-junit4-junit5). Again, I didn't check this thoroughly so let's do the cleanup as a part of the dedicated #1365 task.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/content/WorkflowTest.java
marekhorst: I will stick to `@Disabled` instead of `@DisabledIf` because in the end we want to have it enabled just need to fix the environment.

marekhorst:
Already merged with IIS master branch with [this](https://github.com/openaire/iis/commit/a354afcfc7dd36c98411a987dc7f822a9878611e) squashed commit.

PR #1364, opened 2022-06-29
Closes #1363: Fix jackson library mismatch causing class incompatibility fatal error while importing the graph

### LINKED ISSUE #1363: Fix jackson library mismatch causing class incompatibility fatal error while importing the graph
Running the primary workflow while preparing #1361 pull request revealed the fatal issue causing `import_infospace` phase to fail with the following error message:

`java.io.InvalidClassException: com.fasterxml.jackson.databind.util.StdDateFormat; local class incompatible: stream classdesc serialVersionUID = 2362066295480386373, local class serialVersionUID = 8976348737753383270`

which clearly shows incompatibility of jackson classes caused by misaligned jackson lib versions.

Link to the spark job: http://iis-cdh5-test-gw.ocean.icm.edu.pl:18089/history/application_1614993033499_112964/2/jobs/job/?id=0

It seems the already released `/lib/iis/primary/snapshots/2022-05-06` IIS version is also effected so it should be hot-fixed.

marekhorst:
This is quite important PR because if fixes fatal error in InfoSpace graph importer phase.

Excluding conflicting jackson dependencies from dhp-schemas.

### DIFF
--- a/iis-common/pom.xml
+++ b/iis-common/pom.xml
@@ -35,6 +35,23 @@
         <dependency>
             <groupId>eu.dnetlib.dhp</groupId>
             <artifactId>dhp-schemas</artifactId>
+            <exclusions>
+                <exclusion>
+                    <!-- conflicting versions -->
+                    <groupId>com.fasterxml.jackson.core</groupId>
+                    <artifactId>jackson-core</artifactId>
+                </exclusion>
+                <exclusion>
+                    <!-- conflicting versions -->
+                    <groupId>com.fasterxml.jackson.core</groupId>
+                    <artifactId>jackson-databind</artifactId>
+                </exclusion>
+                <exclusion>
+                    <!-- conflicting versions -->
+                    <groupId>com.fasterxml.jackson.core</groupId>
+                    <artifactId>jackson-annotations</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>


marekhorst:
The tests run after excluding those deps proved the issue was resolved.

mpol (approved):
If it works it works, although I would be happier if I understood better what exactly is going on.

mpol:
To expand a bit on what I understand and don't:

The error is caused by Java deserialization with a different version of a class than was serialized. It is in the context of counting element in some RRD, so shouldn't the serialization and deserialization happen in the same Spark application?

In any case http://iis-cdh5-test-gw.ocean.icm.edu.pl:18089/history/application_1614993033499_112964/2/environment/ shows three versions of jackson-databind on the classpath: 2.2.3 from CDH, and 2.10.3 and 2.9.6 from the application.
So after removing one we will still have two on the classpath. Although if it worked before, it will probably continue working. 

Looking at the serialVersionUID in the error the serialization happened with 2.9.6 and deserialization is attempted with 2.10.3. And I'm not exactly sure where the 2.9.6 is even coming from. Which dependency actually pulls it in?

marekhorst:
Should we consider this as a quick fix (just to let the provisioning round going) and take a closer look later to understand why removing the `2.10.3` version of `jackson-databind` while keeping `2.9.6` does solve this issue?

marekhorst:
OK, I just saw the approval. So I will carry on with the merge.

PR #1374, opened 2022-10-25
Nwo fix separators

johnfouf:
update NWO mining/handle cases with dashes and dots in the project id


marekhorst:
Already merged with 6c6573be381e670b1f61ef05d862197a3d50987b commit.

PR #1383, opened 2022-12-13
solve expected string or buffer error

johnfouf:


marekhorst:
It was integreated with this squashed commit: https://github.com/openaire/iis/commit/56e6a8fc1e3d7d66e37952ade7661c57a4259eb0 in the end.

PR #1387, opened 2022-12-16
add 3 platforms to embrc mining

johnfouf:


marekhorst:
This was used for testing only and the final shape of EMBRC mining was integrated with this PR:

https://github.com/openaire/iis/pull/1473

PR #1400, opened 2023-05-15
Update communitiesextract.sql

johnfouf:
Add INSPIRED mining


marekhorst:
Merged with a dedicated branch: https://github.com/openaire/iis/compare/marekhorst_1403_integrate_INSPIRED_RIs_with_communities_reference_extraction
and supplemented with a follow-up fix. Let's continue our work on that branch.

PR #1411, opened 2023-06-06
Closes #1410: Introduce missing integration test cases for funders still not covered with tests

### LINKED ISSUE #1410: Introduce missing integration test cases for funders still not covered with tests
While analyzing the missing ARC links case (#1395) we also get to know ARC funder was not covered with integration tests. Apparently this was the case for the funders integrated in the first batch of mining algos when adding the integration test case was not part of the mining module integration process.

After further inspection it turned out there were also [other funders](https://docs.google.com/spreadsheets/d/12bsSvXHC7qKmBj3wiIc9TZBCo9p7PK_8TzXnCAzRZsc/edit#gid=0) which should be covered, namely:

* FCT
* HRZZ
* MZOS
* NHMRC
* NSF
* SFI

We should add all the missing test cases to make sure we spot potential regression in mining algorithms accuracy.

marekhorst:
The projects mining integration tests suite was supplemented with the following 6 missing funder cases: FCT, HRZZ, MZOS, NHMRC, NSF, SFI.

### DIFF
--- a/iis-wf/iis-wf-referenceextraction/src/test/resources/eu/dnetlib/iis/wf/referenceextraction/project/data/document_to_project.json
+++ b/iis-wf/iis-wf-referenceextraction/src/test/resources/eu/dnetlib/iis/wf/referenceextraction/project/data/document_to_project.json
@@ -386,7 +386,7 @@
   "documentId": "50|core_ac_uk__::2fea53c390909c2640b2d1a94d53c0a7",
   "projectId": "40|inca________::1e5e62235d094afd01cd56e65112fc63",
   "confidenceLevel": 0.8,
-  "textsnippet": "This work was supported by INCA (plbio 2010-216 and INCa-DGOS-Inserm 6046)"
+  "textsnippet": "This work was supported by INCA (plbio 2010-216 and INCa-DGOS-Inserm 6046) "
 }
 {
   "documentId": "50|arXiv_______::a343cdcd534d696dd93c7ee9d78b9be7",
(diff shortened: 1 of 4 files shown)


iis-wf/iis-wf-referenceextraction/src/test/resources/eu/dnetlib/iis/wf/referenceextraction/project/data/document_to_project.json
```diff
@@ -386,7 +386,7 @@
   "documentId": "50|core_ac_uk__::2fea53c390909c2640b2d1a94d53c0a7",
   "projectId": "40|inca________::1e5e62235d094afd01cd56e65112fc63",
   "confidenceLevel": 0.8,
-  "textsnippet": "This work was supported by INCA (plbio 2010-216 and INCa-DGOS-Inserm 6046)"
+  "textsnippet": "This work was supported by INCA (plbio 2010-216 and INCa-DGOS-Inserm 6046) "
```
marekhorst, line 389: This is quite strange, and noticed this already before: sometimes when I change the set of funders covered with integration tests I deal with the cases as the one highlighted here. White character (space in this case) either shows up or is removed from the end of the `textsnippet` field. It does not happen when the set of input projects/documents remains unchanged, only when I alter it.

@johnfouf do you have any clue what could be the reason for that? It is not super crucial, it is related to `textsnippet` auxiliary field, but it makes every iteration of changing the integration tests set more cumbersome because test fails, I need to find the reason of the failure and quite often it comes down to either removing or adding space character at the end of the `textsnippet` line. Could we e.g. truncate this field content to get rid of this white character related indeterminacy?

PR #1414, opened 2023-07-10
Closes #1407: Do not create an inverse relation when exporting EOSC services mining outcome

### LINKED ISSUE #1407: Do not create an inverse relation when exporting EOSC services mining outcome
Inverse relation, having the same `relClass` and `relType` just `source` and `target` fields swapped, is redundant and doesn't need to be created as a separate action so we could simplify the exporter module.

Originally mentioned in: https://support.openaire.eu/issues/6354#note-73

marekhorst:


mpol (approved):
LGTM

Issue #1415, opened 2023-07-12
Remove avro.schema.input.key redundant properties from map-reduce actions


marekhorst:
While refactoring the `iis-wf-export-actionmanager` code in #1406 it turned out one of `avro.schema.input.key` properties in `sequencefile` workflow.xml file was improperly set in one of the previous commits thus being ineffective:

https://github.com/openaire/iis/blob/184a5154d271f8c9589618978ba47d2b2cf50856/iis-wf/iis-wf-export-actionmanager/src/main/resources/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/oozie_app/workflow.xml#L492C75-L492C75

`eu.dnetlib.iis.referenceextraction.dataset.schemas.DocumentToDatasource` schema was not handled by the `generate-schema` action prior to `exporter-document-to-datasource` execution. 

Even though integration test did not fail proving this property is redundant and can be simply removed.

After identifying this issue and removing `avro.schema.input.key` propertly completely from the exporter actions the integration test did not fail proving once again this property does not need to be defined.

We should drop all the `avro.schema.input.key` properties from all exporter actions along with the `generate-schema` action which also is not required anymore.

We should also check other map-reduce cases in other modules involving the `generate-schema` pattern.

marekhorst:
`avro.schema.input.key` was definitely required for `map-reduce` actions at some point in the past but apparently this is not the case anymore.

PR #1416, opened 2023-07-14
Closes #1406: Align classes and resources naming around the EOSC services mining

### LINKED ISSUE #1406: Align classes and resources naming around the EOSC services mining
This should be a part of #1404 but since we are in a rush and should not postpone new IIS version release let's cover this activity in a dedicated, follow-up task.

After changing some datasource references to services among exported metadata we should probably align classes names including `Datasource` chunk in their names (to be replaced with `Service)`. This also refers to `document_to_dataset_` resources and possibly (to be considered if we should keep it generic) `DocumentToDatasource.avdl` name and `datasourceId` field name.

Classes in `iis-wf-referenceextraction` and `iis-wf-export` modules should be inspected.

marekhorst:
Renaming most of the `datasource` occurrences with `service` phrase to better reflect the scope of the EOSC services mining. Keeping `datasource` naming in the `iis-wf-importer` module where it is more appropriate.

In addition some of the `eoscservice` references were also replaced with `service` to further align the naming. 

Not all of them were covered because we have already agreed on the contract between the IIS and the provisioning orchestrator (DNet part of the system) regarding the property names so the following names remain unchanged:
* `active_referenceextraction_eoscservice`
* `import_infospace_eligible_service_collectedfrom_datasourceid`


marekhorst:
Unfortunately the integreation tests suite run on the branch got frozen due to the lack of free space on the `iis-ci-test` machine so we don't have a confirmation all the integration tests succeeded. 

I am about to rerun those tests although we reached the `iis-wf-referenceextraction` phase so only the `iis-wf-primary` was unchecked. This test was run manually in isolation and it succeeded. So to my understanding we are OK.

iis-schemas/src/main/avro/eu/dnetlib/iis/referenceextraction/service/DocumentToService.avdl
```diff
@@ -1,12 +1,12 @@
-@namespace("eu.dnetlib.iis.referenceextraction.datasource.schemas")
+@namespace("eu.dnetlib.iis.referenceextraction.service.schemas")
 protocol IIS{
 
-    record DocumentToDatasource {
+    record DocumentToService {
 //      document identifier, foreign key: DocumentWithBasicMetadata.id ("document basic metadata" data store)
         string documentId;
 //      identifier of the referenced datasource,
```
mpol, line 7: the comment should be updated too

iis-wf/iis-wf-export-actionmanager/src/main/resources/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/oozie_app/workflow.xml
```diff
@@ -291,6 +291,7 @@
             </main-class>
             <arg>eu.dnetlib.iis.referenceextraction.project.schemas.DocumentToProject</arg>
             <arg>eu.dnetlib.iis.referenceextraction.dataset.schemas.DocumentToDataSet</arg>
+            <arg>eu.dnetlib.iis.referenceextraction.service.schemas.DocumentToService</arg>
```
mpol, line 294: How it worked before without DocumentToDatasource here?

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/DocumentToServiceActionBuilderModuleFactory.java
mpol, line 1: createAction javadoc should be updated

mpol (approved):
Two missed changes in comments and one question, otherwise LGTM.

iis-wf/iis-wf-export-actionmanager/src/main/resources/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/oozie_app/workflow.xml
marekhorst: Apparently this was required in one of the previous versions of CDH, when this `generate-schema` approach was introduced, but then it turned out to be redundant at some point.

Last week I have created a dedicated issue for this, #1415, as a follow-up task I will probably pick up in the next sprint.

Issue #1417, opened 2023-07-18
Introduce missing integration test for avro to protobuf workflow


marekhorst:
When working on #1415 it turned out `avro_to_protobuf` workflow is not covered with integration test which should be implemented.

PR #1418, opened 2023-07-18
Closes #1415: Remove avro.schema.input.key redundant properties from map-reduce actions

### LINKED ISSUE #1415: Remove avro.schema.input.key redundant properties from map-reduce actions

marekhorst:
Removing all `avro.schema.input.key` references in all workflows where it was defined. 

Dropping `generate-schema` action whenever the class defined as a value of removed `avro.schema.input.key` property was the only one covered with schema generation action.

For almost all of the cases integration tests proved the introduced changes did not affect the job in any way. The only case not covered with test was checked by the manual run of the documents similarity workflow (which involved `avro_to_protobuf` conversion).

I have created another task #1417 on creating missing integration test for `avro_to_protobuf` which due to some strange reason was not covered with integration test yet.

### DIFF
--- a/iis-common/src/main/resources/eu/dnetlib/iis/common/protobuf/converter/avro_to_protobuf/oozie_app/workflow.xml
+++ b/iis-common/src/main/resources/eu/dnetlib/iis/common/protobuf/converter/avro_to_protobuf/oozie_app/workflow.xml
@@ -14,10 +14,6 @@
             <name>param_converter_class</name>
             <description>a class implementing AvroToProtoBufConverter</description>
         </property>
-        <property>
-            <name>param_avro_input_class</name>
-            <description>Avro input class</description>
-        </property>
 	</parameters>
     
     <global>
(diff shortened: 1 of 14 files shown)


mpol (approved):
LGTM

PR #1419, opened 2023-07-25
Update projects.sql

johnfouf:
Add HFRI mining


marekhorst:
Replaced with https://github.com/openaire/iis/pull/1421

PR #1425, opened 2023-08-03
Closes #1417: Introduce missing integration test for avro to protobuf workflow

### LINKED ISSUE #1417: Introduce missing integration test for avro to protobuf workflow

marekhorst:
This PR introduces missing test verifying correctness of agro -> protobuf transformation workflow.

The major difficulty was to implement a testing consumer verifying whether produced protobuf outcome is built according to predefined expectations. Usually we rely on a simple java class (`Process` in the IIS nomenclature) but it turned out to be rather cumbersome to veriify protocol buffer record this way because a lot of new protobuf related code seemed to be required to be written in order to introduce protobuf-compatible `TestingConsumer`. So I decided to implement less generic, spark based but very simple `PersonProtoConsumer` which does the job.

### DIFF
--- a/iis-wf/iis-wf-documentssimilarity/pom.xml
+++ b/iis-wf/iis-wf-documentssimilarity/pom.xml
@@ -51,6 +51,18 @@
             <classifier>hadoop2</classifier>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-core_2.11</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-sql_2.11</artifactId>
+            <scope>test</scope>
+        </dependency>
+
         <!-- It seems that this dependency is needed, if it is not here,
         running Oozie tests of Avro-based map-reduce ends up with
         "java.lang.NoClassDefFoundError" exception and statement that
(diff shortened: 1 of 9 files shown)


iis-wf/iis-wf-documentssimilarity/pom.xml
```diff
@@ -86,5 +98,48 @@
...
+                        <configuration>
+                            <sources>
+                                <source>${project.build.directory}/generated-test-sources/proto/</source>
+                            </sources>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+		</plugins>
+    </build>
```
mpol, line 143: This can be simplified, avoiding the build helper plugin.

```suggestion
    
    <build>
        <plugins>
            <plugin>
                <groupId>com.github.os72</groupId>
                <artifactId>protoc-jar-maven-plugin</artifactId>
                <executions>
                    <execution>
                        <id>generate-protobuf-test-sources</id>
                        <phase>generate-test-sources</phase>
                        <goals>
                            <goal>run</goal>
                        </goals>
                        <configuration>
                            <protocVersion>2.5.0</protocVersion>
                            <inputDirectories>
                                <include>src/test/protobuf</include>
                            </inputDirectories>
                            <addSources>test</addSources>
                            <outputDirectorySuffix>proto</outputDirectorySuffix>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
		</plugins>
    </build>
```

mpol (approved):
There is a simplification possible in the build, otherwise looks good.

Issue #1428, opened 2023-08-18
Fuzzy citation matching input and output transformers do not remove output dir at startup


marekhorst:
This may result in input or output transformer spark job failure when reattempting citation matching workflow without prior output directory removal. This is usually the case when reruning citation matching job run in isolation (dev execution mode).

Issue #1427, opened 2023-08-16
Citation matching failed due to an executor exceeding memory limits


marekhorst:
Originally requested in: https://support.openaire.eu/issues/8966

Direct citation matching phase failed on BETA with:

```
Job aborted due to stage failure: Task 51 in stage 5.0 failed 4 times, most recent failure: Lost task 51.3 in stage 5.0 (TID 37918, eos-m2-sn05.ocean.icm.edu.pl, executor 22): ExecutorLostFailure (executor 22 exited caused by one of the running tasks) Reason: Container killed by YARN for exceeding memory limits. 8.1 GB of 8 GB physical memory used. Consider boosting spark.yarn.executor.memoryOverhead or disabling yarn.nodemanager.vmem-check-enabled because of YARN-4714.
```

Already deployed IIS workflow was hotpatched but we should provide a long term solution by preparing direct_citationmatching and primary_processing workflows to allow executor memory overhead to be adjusted at runtime. 

After we find out the new memory related config it should be committed to the gitlab repo at ICM where we keep the config-default.xml file template.

marekhorst:
After hotpatching direct citation matching one of the subsequent fuzzy citation matching steps (transformation) has failed with similar error:
```
Container killed by YARN for exceeding memory limits. 11.3 GB of 11 GB physical memory used. Consider boosting spark.yarn.executor.memoryOverhead.
```
which means more than one module needs to have memory configuration adjusted.

PR #1429, opened 2023-09-06
Introduce two citation matching related fixes for issues discovered during the last provision round on BETA

marekhorst:
Namely:
* #1427 extending executor memory overhead by declaring an explicit value (default value to be set in an env config file addressed [here](https://git.icm.edu.pl/openaire/iis-deployment/-/merge_requests/15))
* #1428 removing input directory in transformer modules (this is the default approach for all spark modules somehow omitted in fuzzy citation matching module what causes issues when rerunning citationmatching in isolation)

Usually I will create two separate PRs but those are pretty small fixes so we could cover them with a single PR (will be merged with the master branch with separate commits though).

### DIFF
--- a/iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/processing/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/processing/oozie_app/workflow.xml
@@ -277,6 +277,11 @@
             <description>memory for individual executor</description>
         </property>
         <property>
+            <name>citationmatchingDirectSparkExecutorOverhead</name>
+            <value>2048</value>
+            <description>The amount of off heap memory (in megabytes) to be allocated for the executor</description>
+        </property>
+        <property>
             <name>citationmatchingDirectSparkExecutorCores</name>
             <value>${sparkExecutorCores}</value>
             <description>number of cores used by single executor</description>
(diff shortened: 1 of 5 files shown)


iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/processing/oozie_app/workflow.xml
```diff
@@ -276,6 +276,11 @@
             <value>${sparkExecutorMemory}</value>
             <description>memory for individual executor</description>
         </property>
+        <property>
+            <name>citationmatchingDirectSparkExecutorOverhead</name>
+            <value>2048</value>
```
mpol, line 281: I don't understand why this value has to be specified both here and in the `config-default.template`. And to they have to be always kept in sync?

mpol (approved):
LGTM, with one question.

iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/processing/oozie_app/workflow.xml
marekhorst: This is in line with other `*SparkExecutorOverhead` properties defined in `workflow.xml` file and it does not require keeping in sync values declared in `workflow.xml` file with the ones from the `default-config.xml` files. 

I think this duplication was mostly kept as safety measure: if the value is not declared in `default-config.xml` it will be taken from the `workflow.xml` as a default one. It is quite useful when performing development runs (no `default-config.xml` involved) but I agree it does not look well having some properties defined in two places therefore I have created a dedicated issue to consider duplicates removal: #1430.

PR #1449, opened 2024-03-05
Update communitiesextract.sql - add check for risis concept label

johnfouf:


marekhorst:
Integrated with this commit:
https://github.com/openaire/iis/commit/f2b434c181bdada49683da2734d3aee1c604f517
having description slightly updated.

PR #1451, opened 2024-04-19
Update projects.sql add HRB mining

johnfouf:


marekhorst:
Merged with this squashed commit:

https://github.com/openaire/iis/commit/364f4660801e418b69bbdb6e555c6051c47db418

supplemented with the dedicated HRB test case.

Issue #1466, opened 2024-09-10
Fix the JATS XML parsing for affiliations coming from Springer


marekhorst:
This is kind of a follow up of the #1464 where the affiliation parsing issue was spotted.

Current version of JATS XML hander prepared for EuropePMC does not support complex affiliations element parsing such as:

```
        <aff id="Aff1">
          <label>Aff1</label>
          <institution-wrap>
            <institution-id institution-id-type="GRID">grid.429509.3</institution-id>
            <institution-id institution-id-type="ISNI">0000000404914256</institution-id>
            <institution content-type="org-name">Max-Planck-Institute for Immunobiology</institution>
          </institution-wrap>
          <addr-line content-type="street">Schillhof 5</addr-line>
          <addr-line content-type="postcode">79110</addr-line>
          <addr-line content-type="city">Freiburg</addr-line>
        </aff>        
```
which results in the following raw affiliation text extracted from the XML:
```
grid.429509.3 0000000404914256 Max-Planck-Institute for Immunobiology  Schillhof 5 79110 Freiburg Germany
```

This is clearly wrong because all the chunks of text are glued togheter. 

In EPMC affiliations were simply encoded as:
```
<aff id="I1"><label>1</label>Graduate School of Bioscience and Biotechnology, Tokyo Institute of Technology, Nagatsuta-cho, Midori-ku, Yokohama 226-8501, Japan</aff>
```
which means the elements of affiliation were not tokenized and the raw text was simply defined as a part of the aff element.

Now we should align the importer with the complex affiliation definition coming from Springer. What is more, we should not rely on the tokenization tool coming from CERMINE in order to parse the raw string and identify organization, address, country because it is already identified and marked with an XML tag.

marekhorst:
The fix introduced with this commit:
https://github.com/openaire/iis/commit/a8f5a302877d9241e94d3a6f67c364e19cddc7cf
covered also #1464 even though it was not explicitly mentioned (my mistake) in the commit message.

PR #1472, opened 2024-09-20
Marekhorst 1469 reimplement software mining algorithm to rely on spark while joining mining output with software heritage links

marekhorst:
Some hits that should help in reviewing the code:
* this commit includes changes introduced by Yannis in the `softwarepost.sql` madis script. The script was modified by dropping the joining with SH origins which relied on SQLite database (which does not fit on the datanode anymore). 
* `SoftwareHeritageUrlsMatchingJob` spark job was implemented to take over the joining operation in a more parallelizable fashion. It was covered with both unit and integration tests. It is run after the modified `softwarepost.sql` job is run and takes over the joining operation.
* newly introduced `DocumentToSoftwareUrlPreMatching` class represents a new record type returned by the modified version of the `softwarepost.sql` madis script. It conveys `cleanmatch` which is used by the spark job performing join between the SH origin (`SoftwareHeritageOrigin#url`) and `DocumentToSoftwareUrlPreMatching#cleanmatch`
* `softwareheritage_sqlite_builder` phase is not needed anymore so all the action references were dropped as well as the java code and workflow definition
* all `origins.db` SQLite db occurences in `workflow.xml` definitions were replaced with HDFS path pointing to the avro datastore with `SoftwareHeritageOrigin` records
* whenever spark job cannot join `SoftwareHeritageOrigin` record with `DocumentToSoftwareUrlPreMatching` it returns `null` `SHUrl` instead of an empty string value in the output with the `DocumentToSoftwareUrlWithMeta` records. I did not fix that because both `null` and "" are properly handled down the processing pipe and I don't think "" is better than `null` here.
* I had to use `TaraReferenceExtractionIOUtils` in this context so I made it more generic (name and package wise).


mpol (approved):
LGTM

marekhorst:
Merged with the master branch (after rebase due to #1466).

PR #1500, opened 2024-11-25
Closes #1494: Prepare a command line tool allowing generation of an output of PMC ingestion parsing run on a particular XML file

### LINKED ISSUE #1494: Prepare a command line tool allowing generation of an output of PMC ingestion parsing run on a particular XML file
This tool is required to work effectively on affiliation matching recalibration task.

We need a similar class to `ContentExtractor` which is part of CERMINE and allows easy extraction of metadata from an individual PDF file. We should have a PMC ingestion equivalent working on XML file(s) and producing outcome for analysis.

Since the core of the analysis is affiliation extraction (including parsing and tokenization of the affiliation text) it will be good to have another command line tool accepting affiliation text and extracting affiliation metadata in the same way it is done by the ingester module. This would also speed-up the analysis process and help in finding issues with affiliation texts tokenization.

marekhorst:
This commit brings two utility classes (command line tools) helping in an XML JATS file analysis and affiliation parsing.

`MetadataImporterMain` class is responsible for triggering PMC ingestion from the XML file provided at input and producing the JSON representation of extracted metadata.

`AffiliationMetadataExtractorMain` extracts affiliation metadata directly from the affiliation text provided at input (as a single cmndline parameter so it needs to be wrapped with quotes) and produces JSON representation of parsed affiliation object.

Those are utility classes aimed to be used in debugging process, they are not meant to be used in production environment, so they are defined in `src/test`.

### DIFF
--- a/iis-wf/iis-wf-ingest-pmc/src/test/java/eu/dnetlib/iis/wf/ingest/pmc/metadata/AffiliationMetadataExtractorMain.java
+++ b/iis-wf/iis-wf-ingest-pmc/src/test/java/eu/dnetlib/iis/wf/ingest/pmc/metadata/AffiliationMetadataExtractorMain.java
@@ -1,0 +1,74 @@
+package eu.dnetlib.iis.wf.ingest.pmc.metadata;
+
+import java.io.ByteArrayOutputStream;
+
+import org.apache.avro.io.DatumWriter;
+import org.apache.avro.io.Encoder;
+import org.apache.avro.io.EncoderFactory;
+import org.apache.avro.specific.SpecificDatumWriter;
+import org.apache.commons.lang.StringUtils;
+import org.jdom.Element;
+import org.xml.sax.SAXException;
+
+import eu.dnetlib.iis.common.importer.CermineAffiliation;
+import eu.dnetlib.iis.common.importer.CermineAffiliationBuilder;
+import eu.dnetlib.iis.ingest.pmc.metadata.schemas.Affiliation;
+import pl.edu.icm.cermine.exception.AnalysisException;
+import pl.edu.icm.cermine.exception.TransformationException;
+import pl.edu.icm.cermine.metadata.affiliation.CRFAffiliationParser;
+
+/**
+ * 
+ * Simple command line tool accepting affiliation string, to be provided as a single cmdline argument 
+ * (so needs to be specified in quotes), and producing extracted affiliations metadata.
+ * 
+ * @author mhorst
+ */
+public class AffiliationMetadataExtractorMain {
+
+    private static final String encoding = "utf-8";
+    
+    private static final CermineAffiliationBuilder cermineAffiliationBuilder = new CermineAffiliationBuilder();
+    private static final CermineToIngestAffConverter cermineToIngestAffConverter = new CermineToIngestAffConverter();
+    
+    public static void main(String[] args) throws Exception {
+        if (args.length == 1) {
+            Affiliation aff = buildAffiliationFromText(args[0]);
+            if (aff == null) {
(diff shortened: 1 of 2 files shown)


iis-wf/iis-wf-ingest-pmc/src/test/java/eu/dnetlib/iis/wf/ingest/pmc/metadata/AffiliationMetadataExtractorMain.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+public class AffiliationMetadataExtractorMain {
+
+    private static final String encoding = "utf-8";
+    
+    private static final CermineAffiliationBuilder cermineAffiliationBuilder = new CermineAffiliationBuilder();
+    private static final CermineToIngestAffConverter cermineToIngestAffConverter = new CermineToIngestAffConverter();
+    
+    public static void main(String[] args) throws Exception {
+        if (args.length >= 1) {
```
mpol, line 35: Since it only uses one argument it should check that there is exactly one. It will also help to catch forgotten quoting.

iis-wf/iis-wf-ingest-pmc/src/test/java/eu/dnetlib/iis/wf/ingest/pmc/metadata/MetadataImporterMain.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * This class is responsible for performing metadata extraction out of the JATS XML file which location is provided as input parameter.
+ * @author mhorst
+ */
+public class MetadataImporterMain {
+    
+    private static final String encoding = "utf-8"; 
+    
+    public static void main(String[] args) throws IOException, ParserConfigurationException, SAXException {
+        if (args.length >= 1) {
```
mpol, line 34: Since it only uses one argument it should check that there is exactly one. It will also help to catch forgotten quoting - although it's unlikely to have a path containing whitespace.

iis-wf/iis-wf-ingest-pmc/src/test/java/eu/dnetlib/iis/wf/ingest/pmc/metadata/MetadataImporterMain.java
```diff
@@ -0,0 +1,84 @@
...
+        JatsXmlHandler pmcXmlHandler = new JatsXmlHandler(output);
+        saxParser.parse(new InputSource(new StringReader(xmlInput)), pmcXmlHandler);
+    }
+    
+    /**
+     * Reads file content from the location provided as input parameter.
+     */
+    private static String getFileContent(String location) throws IOException {
+        return new String(Files.readAllBytes(Paths.get(location)), Charset.forName(encoding));
+    }
```
mpol, line 83: Not sure if going `byte[]` -> `String` -> `StringReader` is better than `byte[]` -> `ByteArrayInputStream`. As long as the file encoding is actually UTF-8 both should work.

mpol (approved):
I have one minor comment, and one comment/question but feel free to merge as is.

iis-wf/iis-wf-ingest-pmc/src/test/java/eu/dnetlib/iis/wf/ingest/pmc/metadata/MetadataImporterMain.java
marekhorst: I think both methods were copied from different places what introduced this conversion overhead. I've simplified both methods a little bit in order to stick to simple byte[] representation.

PR #1506, opened 2024-12-16
Closes #1505: Update command line tool responsible for PMC/JATS records parsing to support directory as an input

### LINKED ISSUE #1505: Update command line tool responsible for PMC/JATS records parsing to support directory as an input
Currently the tool introduced in #1494 supports an XML file provided as input path but it will be convenient to provide path pointing to a directory to parse all the files located inside that directory.

Once we are capable of processing multiple files we should, apart from dumping an outcome of parsing to stdout just as it is done currently, write the outcome JSON files to the file system for further processing and analysis. The prduced file names should be prefixed in a way they won't be picked up in the subsequent bulk parsing performed in the same directory.

marekhorst:
This is a simple improvement: tool is meant to accept not only XML file but also a directory of files as input. No recursion is needed in parsing subdirs.

Apart for supporting directory as input it generates JSON files suffixed with `.parsed`.

### DIFF
--- a/iis-wf/iis-wf-ingest-pmc/src/test/java/eu/dnetlib/iis/wf/ingest/pmc/metadata/MetadataImporterMain.java
+++ b/iis-wf/iis-wf-ingest-pmc/src/test/java/eu/dnetlib/iis/wf/ingest/pmc/metadata/MetadataImporterMain.java
@@ -2,6 +2,7 @@ package eu.dnetlib.iis.wf.ingest.pmc.metadata;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
@@ -14,6 +15,7 @@ import org.apache.avro.io.DatumWriter;
 import org.apache.avro.io.Encoder;
 import org.apache.avro.io.EncoderFactory;
 import org.apache.avro.specific.SpecificDatumWriter;
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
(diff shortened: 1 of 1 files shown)


iis-wf/iis-wf-ingest-pmc/src/test/java/eu/dnetlib/iis/wf/ingest/pmc/metadata/MetadataImporterMain.java
```diff
@@ -29,30 +32,52 @@ public class MetadataImporterMain {
     
     private static final String encoding = "utf-8"; 
     
+    private static final String parsedFileSuffix = ".parsed";
```
mpol, line 35: Wouldn't it be convenient for the output files to have ".parsed.json" extension? Eg. for automatic syntax highlighting in editors?

iis-wf/iis-wf-ingest-pmc/src/test/java/eu/dnetlib/iis/wf/ingest/pmc/metadata/MetadataImporterMain.java
```diff
@@ -29,30 +32,52 @@ public class MetadataImporterMain {
...
-            System.out.println(new String(stream.toByteArray(), encoding));
+            File file = new File(args[0]);
+            if (file.isDirectory()) {
+                String[] fileLocs = file.list(new FilenameFilter() {
+                    
+                    @Override
+                    public boolean accept(File dir, String name) {
+                        return !name.endsWith(parsedFileSuffix);
+                    }
+                });
```
mpol, line 51: The filter would be more readable as a lambda expression, I think.

iis-wf/iis-wf-ingest-pmc/src/test/java/eu/dnetlib/iis/wf/ingest/pmc/metadata/MetadataImporterMain.java
```diff
@@ -29,30 +32,52 @@ public class MetadataImporterMain {
...
+        extractMetadata(getFileContent(fileLocation), builder);
+        
+        DatumWriter<ExtractedDocumentMetadata> writer = new SpecificDatumWriter<>(ExtractedDocumentMetadata.class);
+        ByteArrayOutputStream stream = new ByteArrayOutputStream();
+        Encoder jsonEncoder = EncoderFactory.get().jsonEncoder(ExtractedDocumentMetadata.getClassSchema(), stream);
+        writer.write(builder.build(), jsonEncoder);
+        jsonEncoder.flush();
+        String fileContent = new String(stream.toByteArray(), encoding);
+        System.out.println(fileContent);
+        FileUtils.writeStringToFile(new File(fileLocation + parsedFileSuffix), fileContent, encoding);
```
mpol, line 78: Possibly not worth other necessary changes, but we could avoid the `FileUtils` dependency by using standard `Files.write(Path, byte[])`

mpol (approved):
Looks good, with minor comments.

iis-wf/iis-wf-ingest-pmc/src/test/java/eu/dnetlib/iis/wf/ingest/pmc/metadata/MetadataImporterMain.java
marekhorst: Replaced with:
```
String[] fileLocs = file.list((dir, name) -> !name.endsWith(parsedFileSuffix));
```

iis-wf/iis-wf-ingest-pmc/src/test/java/eu/dnetlib/iis/wf/ingest/pmc/metadata/MetadataImporterMain.java
marekhorst: As already discussed, it is a good idea but to get rid of this utility class and dependency we'd need to alter the code outside of this PR's scope. Since this PR does not introduce any new dependency, it just relies on the set of already available deps, we could do this optimization but as a part of a different task.

PR #1509, opened 2024-12-20
Marekhorst 1495 recalibrate affiliation matching confidence level for cases without affiliation country code specified

marekhorst:
This is a pretty simple PR being result of a very long analysis.

It's main purpose is to reduce the strength of a match whenever `CountryCodeLooseMatchVoter` is involved which usually means country code was not explicitly defined for a given affiliation and voter accepts given match. Currently the strength was too high and this PR lowers it a little bit to make it more aligned with the characteristics of the real data ( the percentage of affiliations without country code in our test data is smaller than in the real world data from the OpenAIRE Research Graph: 27% vs 29%).

So I added several test cases without the country code in order to recalibrate the strength a little bit.


mpol (approved):
LGTM

marekhorst:
Closing with caceabeb62f5a5b867651fcd2942ca634f8b5023 commit after rebase and squash.

PR #1524, opened 2025-03-06
Closes #1523: Reintroduce relying on deduplication relations when making content identifiers aligned with graph identifiers

### LINKED ISSUE #1523: Reintroduce relying on deduplication relations when making content identifiers aligned with graph identifiers
This is basically to revert #1393 because we need to rely on the `DeduplicationMappingConverter` again but also bringing back the logic of translating content identifiers based on dedup relations imported from the graph.

3 years ago this was replaced with #1264, which was about relying on the mapping between the original identifiers and persistent identifiers, and since we decided to run the IIS on a non-deduped version of the graph back then we could simply drop the id translation based on the dedup identifiers.

Now we want to be able to run the IIS also on deduped version of the graph which means we need to rely on two layers of id translation for contents:
1) already implemented translation between the original identifiers (as defined in PDF Aggregation System) and persistent identifiers (defined in the graph)
2) "resurrected" translation between the persistent identifiers and deduplicated identifiers whenever given persistent id was deduped

This way we should end up with content identifiers which are fully matchable with deduped graph. When running IIS on the non deduplicated graph the 2nd id translation will not be applied because there will be no `merges` relations in the non-deduped graph.

It was described in detail in redmine:
https://support.openaire.eu/issues/10410
where the investigation process was described in detail.

marekhorst:
InfoSpace importer now relies on two layes of mappings in order to fully support import from both deduplicated and non-deduplicated graph:
* between the original identifiers (as defined in PDF Aggregation System) and persistent identifiers (defined in the graph)
* between the persistent identifiers and deduplicated identifiers whenever given entity with a persistent id was dedupedlicated

The first mapping was in use up until now when IIS was mostly run on non-deduplicated data. The second mapping was reintroduced after it was replaced by the first mapping as a part of git#1264.

EDIT: issue was also reported and described in detail on redmine: https://support.openaire.eu/issues/10410


iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJobUtilsTest.java
```diff
@@ -71,6 +71,65 @@ public void givenResultRecords_whenObjectStoreIdsAreProduced_thenProperIdentifie
...
+        @Test
+        @DisplayName("Deduplication mapping is applied")
+        public void givenDedupMapping_whenApplyDedupMappingOnTop_thenMappingsAreUpdated() {
+            JavaRDD<IdentifierMapping> originalIdMapping = jsc.parallelize(Arrays.asList(
+                    createIdentifierMapping("new1", "original1"),
+                    createIdentifierMapping("new2", "original2")
+            ));
+            JavaRDD<IdentifierMapping> dedupMapping = jsc.parallelize(Arrays.asList(
+                    createIdentifierMapping("dedup1", "new1"),
+                    createIdentifierMapping("dedup2", "new2")
```
mpol, line 102: It would be probably nice to have some actual duplication in the test here.

mpol (approved):
Looks good. I just have one suggestion for the tests.

iis-wf/iis-wf-import/src/test/java/eu/dnetlib/iis/wf/importer/infospace/ImportInformationSpaceJobUtilsTest.java
marekhorst: I changed the tests a little bit in the most recent commit because the logic has changed, more details in https://support.openaire.eu/issues/10410#note-3, and now we are simply merging (union + distinct) `IdentifierMapping` records coming from two sources: originalId->persistentId mapping and deduplication mapping. So we are not applying dedup mapping on top of the original merging anymore, just keeping them next to each other as unrelated.
This is because we don't have to translate persistentIds from the originalId->persistentId because whenever deduplication was applied persistentId are already translated to deduped ids (because they are coming from new representative records) while we still need to include deduplication mappings for all the cases where there was no originalId->persistentId mapping defined. Otherwise they will be left unchanged like for PubMed contents having identifiers bult based on DOIs thus will be potentially misaligned with the deduped graph entries having new representative record elected.

marekhorst:
@mpol I pushed another commit changing (also simplifying at the same time) the way identifier mappings are combined. 

Detailed rationale behind that change is available in https://support.openaire.eu/issues/10410#note-3, I shared short version in my comment above.

It includes also code review minor fixes (addressing unused imports) from the previous code review.

mpol (approved):
Looks good

marekhorst:
It was merged with the IIS master branch after squashing the commits and altering the description:
https://github.com/openaire/iis/commit/a2b38577d0e098c9d85ab70b5eb3c310049b11d4

PR #1537, opened 2025-04-28
Marekhorst 1534 repartition affiliation matching final output before generating reports

marekhorst:
This Pull Reqest solves the problem originally described in [#10546](https://support.openaire.eu/issues/10546) redmine ticket:

> The most recent IIS run in the production environment has failed with the following error:
> ```
> org.apache.spark.SparkException: Error communicating with MapOutputTracker
> ``` 
> caused by the:
> ``` 
> java.util.concurrent.TimeoutException: Futures timed out after [600 seconds]
> ``` 

In the past we've been usually dealing with this kind of issues by increasing the timeout parameter mentioned in the detailed stack trace (`spark.network.timeout`).

This time, after deeper inspection and looking into the data, I decided to introduce data repartitioning because it was quite evident we are dealing with ids data skew. Since we already repartition data just before saving the output of processing I decided to simply move this operation one step earlier to rely on repartitioning data when generating report.

Tests has proven this eliminated the problem because the report was successfully generated. 

Apart from that, in order to be on a safe side and reduce the probability of the algorithm fatal failure, I doubled the `sparkNetworkTimeout` value. During multiple runs relying on the repartitioning solution mentioned above there were still individual task timeouts (non-fatal this time) reported in the logs.

### DIFF
--- a/iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/write/IisAffMatchResultWriter.java
+++ b/iis-wf/iis-wf-affmatching/src/main/java/eu/dnetlib/iis/wf/affmatching/write/IisAffMatchResultWriter.java
@@ -59,13 +59,12 @@ public class IisAffMatchResultWriter implements AffMatchResultWriter {
                 .reduceByKey((match1, match2) -> matchedOrgStrengthRecalculation.recalculateStrength(match1, match2))
                 .values();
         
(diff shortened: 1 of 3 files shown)


marekhorst:
It got merged with the following squashed commit:

https://github.com/openaire/iis/commit/1244b0f5e01dbd8c640b4bf027b2927ee71f0c27

PR #1538, opened 2025-04-30
Rename RPF to RIF

johnfouf:
More edit is needed here, some new projects have a new pattern.  Will be done in a future commit.


harry-di:
This is high priority due to the Cyprus use case deadline, end of May.

marekhorst:
Integrated with this squashed commit supplemented with a dedicated test case:

https://github.com/openaire/iis/commit/ec8cc6a7ccb994fd3f735374831de1f53258391d

PR #1539, opened 2025-04-30
Marekhorst 1533 introduce caching for the fuzzy citation matching algorithm

marekhorst:
Here is the brief description of introduced changes:
* making `CacheStorage` utils more generic (it was used for storing text records so far), moving `LockManagerFactoryMock` to commons
* modifying `CitationMatchingInputTransformerJob` to handle caching by filtering out citation matching input (dropping records already found in cache from input). It does not remove whole records if found in cache because they are also candidates for matching other bibrefs, it just removes bibrefs for cached records since they were already matched in one of the previous runs.
* modifying `CitationMatchingOutputTransformerJob` to handle caching by
    * storing entries in cache for newly introduced publications published before `currentYear - cacheOlderThanXYears`. We cache also the unmatched bibrefs because we assume we won't find the match in the future runs anyway (hardly possible to get old publication in the graph).
    * retrieving already cached entries referencing set of publications provided at citation matching input (so not retrieving the irrelevant ones) and merging the output with currently matched references
* increasing `citationMatchingSparkNetworkTimeout` from `600s` to `1200s`, just in case, to prevent the job from failing (we were struggling with the timeout in the past in citation and affiliation matching algorithms and we really don't want to deal with fatal failures) 
* adding test cases involving cache utilization
* defining required parameters in uber workflows (mostly in the IIS primary wfs)


iis-wf/iis-wf-citationmatching/src/main/java/eu/dnetlib/iis/wf/citationmatching/input/DocumentToCitationDocumentConverter.java
```diff
@@ -71,7 +73,7 @@ private List<ReferenceMetadata> convertReferences(List<eu.dnetlib.iis.metadataex
         }
```
mpol, line 73: I would rather check `discardAllReferences` here.

iis-wf/iis-wf-citationmatching/src/main/java/eu/dnetlib/iis/wf/citationmatching/output/CitationMatchingOutputTransformerJob.java
```diff
@@ -33,30 +50,78 @@ public class CitationMatchingOutputTransformerJob {
...
+            JavaRDD<ExtractedDocumentMetadataMergedWithOriginal> inputDocuments = avroLoader.loadJavaRDD(sc,
+                    params.inputMetadata, ExtractedDocumentMetadataMergedWithOriginal.class);
+            JavaPairRDD<CharSequence, Integer> inputDocumentsIdtoYear = inputDocuments.mapToPair(x -> new Tuple2<CharSequence, Integer>(x.getId(), x.getYear())); 
+            // caching because this RDD is used for filtering data for both caching and generating final output
+            inputDocumentsIdtoYear.cache();
+
+            // caching only a subset of matched citations coming from a paper published earlier that X years ago
+            // and only the ones which were not cached before (required to cache also entries with no matches which will
+            // not be a part of transformedMatchedCitations but we should avoid incuding them in a subsequent run)
+            int publicationYearThreadshold = Year.now().getValue() - params.cacheOlderThanXYears;
```
mpol, line 101: This requires use of absurd future year values in testing.
But I'm not sure if passing the current year from outside just for testing makes sense, maybe this is good enough.

iis-wf/iis-wf-citationmatching/src/main/java/eu/dnetlib/iis/wf/citationmatching/input/CitationMatchingInputTransformerJob.java
```diff
@@ -51,36 +56,44 @@ public static void main(String[] args) throws InterruptedException, IOException
...
+            
+            JavaPairRDD<CharSequence, ExtractedDocumentMetadataMergedWithOriginal> pairedDocuments = inputDocuments.mapToPair(doc -> new Tuple2<>(doc.getId(), doc));
+            
+            JavaPairRDD<CharSequence, Tuple2<Tuple2<ExtractedDocumentMetadataMergedWithOriginal, Optional<Iterable<Integer>>>, Optional<Boolean>>> inputDocumentsJoinedWithMatchedCitations = pairedDocuments
+                    .leftOuterJoin(groupedMatchedCitations).leftOuterJoin(groupedCachedCitations);
+            
+            JavaRDD<DocumentMetadata> documents = inputDocumentsJoinedWithMatchedCitations
+                    .map(x -> documentToCitationDocumentConverter.convert(x._2._1._1,
+                            x._2._1._2.isPresent() ? Sets.newHashSet(x._2._1._2.get()) : Collections.emptySet(),
+                            x._2._2.isPresent() ? x._2._2.get() : false));
```
mpol, line 94: At this level of nesting tuples without pattern matching are really quite unreadable. Maybe at least introduce local variables in the lambda for `x._2._1._2` and `x._2._2`?

marekhorst: Yeap, I hate that nesting with artificial `_n` field identification too. I guess it is worth to sacrifice compactness of the code for the increased readability. I got back to this code after several weeks and even though I wrote it by myself it was really difficult for me to "decode" what is really going on here.

I just changed it with:
```
            JavaRDD<DocumentMetadata> documents = inputDocumentsJoinedWithMatchedCitations
                    .map(x -> {
                        ExtractedDocumentMetadataMergedWithOriginal inputDocumentMetadata = x._2._1._1;
                        Optional<Iterable<Integer>> alreadyMatchedReferencePositions = x._2._1._2;
                        Optional<Boolean> isDocumentAlreadyCached = x._2._2;

                        return documentToCitationDocumentConverter.convert(inputDocumentMetadata,
                                alreadyMatchedReferencePositions.isPresent()
                                        ? Sets.newHashSet(alreadyMatchedReferencePositions.get())
                                        : Collections.emptySet(),
                                isDocumentAlreadyCached.isPresent() ? isDocumentAlreadyCached.get() : false);
                    });
```

marekhorst:
This feature was just merged with the IIS master branch with this rebased and squashed commit:

https://github.com/openaire/iis/commit/4f9d87c92020d8c914e7b26bf8234c7e7a9f7fd1

PR #1545, opened 2025-05-28
Closes #1544: Reduce the number of files generated by the citation matching algorithm

### LINKED ISSUE #1544: Reduce the number of files generated by the citation matching algorithm
This is kind of similar issue to #1441 and #1442 in a way the outcome generated by the citationmatching algorithm is split into over 32k files (as of today: 33985), each file size is below 2MB, and this will result in the following type of errors:

```
 SocketException: Too many open files
```
which could be potentially generated by the next module down the processing chain.

We should adjust the number of generated files by citation matching algorithm and reduce it to a reasonably low number.

marekhorst:
Reducing the number of generated files by repartitioning, 1000 files is generated by default.

Fixing parameter names to avoid confusion: `citationmatchingCacheNumberOfEmittedFiles` was defined before which was pretty ambiguous knowing we are supporting now two parameters: one related to the number of files generated by the algorithm at output and the other one related to the number of files generated in cache.


marekhorst:
Merged with the IIS master branch with this cherry-picked commit:
https://github.com/openaire/iis/commit/1fd41cf711f13c312c20b2e8b749dc5a7beb6a10
The branch where  the feature was developed based on another citationmatching related feature branch which was merged with a master branch after the squash.

PR #1561, opened 2025-09-04
Closes #1560: Introduce extractedBy field in the ExtractedDocumentMetadata record indicating its origin

### LINKED ISSUE #1560: Introduce extractedBy field in the ExtractedDocumentMetadata record indicating its origin
We should introduce provenance field (e.g. `extractedBy`) in the metadata extraction record in order to indicate the tool responsible for generating given metadata record (CERMINE or GROBID). We could make it more fine-grained and apart from the metadata extraction tool name we could define the tool version for a better control.

This will be extremely useful for updating the already built metadata extraction cache currently having all records originating from CERMINE. Once we introduce new records provided by Grobid or any other metadata extraction tool, potentially fixing the existing ones, we should mark those records appropriately. This way we would know what exact tool was responsible for metadata extraction and we could exclude some records provided by given tool from further updates.

marekhorst:
Introducing `extractedBy` field in the `ExtractedDocumentMetadata` record. Setting `CERMINE` value in the base metadata extraction module relying on CERMINE. Aligning full set of integration tests.


mpol (approved):
LGTM. Although I have a small doubt about setting extractedBy CERMINE for the empty records.

PR #1572, opened 2025-10-03
Closes #1569: Both sparkExecutorOverhead and sparkExecutorMemory values should be declared in the same resource

### LINKED ISSUE #1569: Both sparkExecutorOverhead and sparkExecutorMemory values should be declared in the same resource
This is #1566 follow-up. 

We should align on the location where both `sparkExecutorMemory` and `sparkExecutorOverhead` properties are defined. Currently the `sparkExecutorOverhead` value is declared within the `workflow.xml` definition file while the `sparkExecutorMemory` value is declared in the external `default-config.xml` resource. 

Since both are related (and the default value of `sparkExecutorOverhead` is `0.1*sparkExecutorMemory` with a minimum of `384`) we should keep their values declaration together.

This should introduced for all the spark modules where both `sparkExecutorMemory` and `sparkExecutorOverhead` are explicitly defined.

marekhorst:
The following properties values were moved to an external `config-default.xml` file:
* `infospaceImportSparkExecutorOverhead`
* `researchInitiativeReferenceExtractionSparkExecutorOverhead`
* `affiliationmatchingSparkDriverOverhead`
* `citationmatchingDirectSparkExecutorOverhead`
* `citationmatchingFuzzySparkDriverOverhead`
* `citationmatchingFuzzySparkExecutorOverhead`

Default values were introduced in the integration test suite whenever missing.

The following commit introduces new values in an external `config-default.xml` file:
https://git.icm.edu.pl/openaire/iis-deployment/-/commit/97528b7c3e195f32ac48830274d8a4a9a6f8b917

### DIFF
--- a/iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/import/oozie_app/workflow.xml
+++ b/iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/import/oozie_app/workflow.xml
@@ -30,7 +30,6 @@
         </property>
         <property>
             <name>infospaceImportSparkExecutorOverhead</name>
-            <value>2048</value>
             <description>The amount of off heap memory (in megabytes) to be allocated for the infospace importer executor</description>
         </property>
 		<property>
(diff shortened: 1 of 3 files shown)


mpol (approved):
Looks good.
It's interesting that some of the overhead values were already present in the external file. And also that both the driver overheads moved to the external configuration are actually lower than the default would be.

Issue #1596, opened 2026-05-21
Rewrite PIG transformers to spark and make them executable in the k8s environment


marekhorst:
The goal of this task is to rewrite the logic defined in PIG scripts into spark (v 4.1.1) java classes.
Apart from that, test workflows defined in oozie should be ported to airflow workflows at:
https://github.com/openaire/iis-workflows
EDIT: here is the relevant PR: https://github.com/openaire/iis-workflows/pull/2

marekhorst:
This feature was successfully integrated with the [k8s branch](https://github.com/openaire/iis/tree/marekhorst_1588_make_the_IIS_compliant_with_spark4_and_kubernetes_environment_APPROVED).

Issue #1588, opened 2026-02-26
Make the IIS compliant with spark4 and kubernetes environment


marekhorst:
Originally requested in redmine: https://support.openaire.eu/issues/11233.

The major goal of this task is to make it possible to run the IIS on the kubernetes cluster relying on spark 4.1.1 verison.

This is obviously a huge overhaul and major change on many layers involving:
* replacing oozie workflows with airflow
* rewriting spark2.4 jobs to spark4.1
* replacing PIG scripts with the transformations implemented in spark4
* replacing plain M/R jobs with the jobs implemented in spark4
* changing the way Madis scripts are integrated with the IIS

among others.

PR #1597, opened 2026-05-21
Marekhorst 1596 rewrite pig transformers to spark

marekhorst:
This PR introduces the following changes:
* pom.xml update to build uber-jar package with test resources which is required to run integration tests in airflow (test workflows are introduced in the https://github.com/openaire/iis-workflows repo)
* new spark classes realizing the logic which was so far encoded in PIG scirpts which are meant to be abandoned in the k8s environment
* test resources updates - required due to a minor contract changes introduced by the new spark implementation (e.g. `null`s returned instead of empty arrays for fields which are defined as null-friendly)


marekhorst:
The final commit merged into the target branch should not only be squashed but also the commit title should have #1588 replaced with #1596 to relate to the proper issue.

marekhorst:
The integration tests defined as ariflow workflows were introduced in:
https://github.com/openaire/iis-workflows/tree/main/tests

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/skip_extracted/SkipExtractedJob.java
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Spark job that produces two filtered datasets:
+ * <ul>
+ *   <li>outputDocumentContent: DocumentContentUrl records whose id is NOT present in the
+ *       input metadata (i.e. not yet extracted).</li>
+ *   <li>outputDocumentMeta: ExtractedDocumentMetadata records whose publicationTypeName is
+ *       null or not "EMPTY", and whose id IS present in the input document content.</li>
+ * </ul>
+ *
+ * Replaces the PIG-based skip_extracted transformer workflow step.
```
mpol: Not sure if such historical information is useful to keep in code comments. (This is true for basically all the class comments in all the files.)

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/skip_extracted/SkipExtractedJob.java
```diff
@@ -0,0 +1,98 @@
...
+import pl.edu.icm.sparkutils.avro.SparkAvroSaver;
+import scala.Tuple2;
+
+/**
+ * Spark job that produces two filtered datasets:
+ * <ul>
+ *   <li>outputDocumentContent: DocumentContentUrl records whose id is NOT present in the
+ *       input metadata (i.e. not yet extracted).</li>
+ *   <li>outputDocumentMeta: ExtractedDocumentMetadata records whose publicationTypeName is
+ *       null or not "EMPTY", and whose id IS present in the input document content.</li>
```
mpol: First, the implementation checks for "$EMPTY$" not just "EMPTY". Second, while the comment is technically true with the 'or' clause (as in: matches the current implementation) it is confusing as to the intent. If something is null it's definitely not "$EMPTY$", but it's a bit weird that they are opposite and not equivalent here.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/idreplacer/IdReplacerJob.java
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Performs a left join of {@code data} with {@code mapping} on {@code fieldToReplace},
+     * replacing that field's value with {@code newId} from the mapping. Records with no
+     * mapping entry retain their original field value (matching IdReplacerUDF behaviour).
+     */
+    private static Dataset<Row> replaceId(Dataset<Row> data, Schema schema,
+            Dataset<Row> mapping, String fieldToReplace) {
+        Dataset<Row> renamedMapping = mapping.select(
+                col("originalId").alias("_orig"),
+                col("newId").alias("_new"));
```
mpol: Using the `withColumnsRenamed` method could be more readable.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/idreplacer/IdReplacerJob.java
```diff
@@ -0,0 +1,104 @@
...
+            new AvroDataFrameWriter(result).write(params.output, schema);
+        }
+    }
+
+    //------------------------ PRIVATE --------------------------
+
+    /**
+     * Performs a left join of {@code data} with {@code mapping} on {@code fieldToReplace},
+     * replacing that field's value with {@code newId} from the mapping. Records with no
+     * mapping entry retain their original field value (matching IdReplacerUDF behaviour).
```
mpol: "(matching IdReplacerUDF behaviour)" is again historical information that has probably no place here.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/checksum/postprocessing/meta/ChecksumMetaPostprocessingJob.java
```diff
@@ -0,0 +1,85 @@
...
+            HdfsUtils.remove(sc.hadoopConfiguration(), params.output);
+
+            JavaRDD<ExtractedDocumentMetadata> meta = avroLoader.loadJavaRDD(sc,
+                    params.inputExtractedDocumentMetadata, ExtractedDocumentMetadata.class);
+            JavaRDD<DocumentContentUrl> dcu = avroLoader.loadJavaRDD(sc,
+                    params.inputDocumentContentUrl, DocumentContentUrl.class);
+
+            // Build a mapping: checksum -> original document id
+            // After preprocessing, dcu.id IS the checksum, so we use dcu.contentChecksum
+            // but to join with meta.id we use dcu.contentChecksum as the key
```
mpol: This comment is hard to understand and seems confused.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/checksum/postprocessing/fault/ChecksumFaultPostprocessingJob.java
```diff
@@ -0,0 +1,81 @@
...
+import scala.Tuple2;
+
+/**
+ * Spark job that remaps the inputObjectId in Fault records from a contentChecksum back to the
+ * original document id. Only DocumentContentUrl records that have a non-null contentChecksum
+ * are considered. The fault's inputObjectId is expected to match the DCU's contentChecksum.
+ *
+ * Replaces the PIG-based checksum/postprocessing/fault transformer workflow step.
+ */
+public class ChecksumFaultPostprocessingJob {
```
mpol, line 27: This should be merged with `ChecksumMetaPostprocessingJob` - but probably preferably as part of the possible followup task for reducing the number of separate jobs.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/checksum/preprocessing/ChecksumPreprocessingJob.java
```diff
@@ -0,0 +1,81 @@
...
+
+            // Replace id with contentChecksum
+            JavaRDD<DocumentContentUrl> output = dedupedPairs.map(pair ->
+                    DocumentContentUrl.newBuilder()
+                            .setId(pair._2.getContentChecksum())
+                            .setUrl(pair._2.getUrl())
+                            .setMimeType(pair._2.getMimeType())
+                            .setContentChecksum(pair._2.getContentChecksum())
+                            .setContentSizeKB(pair._2.getContentSizeKB())
+                            .build());
```
mpol: Wouldn't `DocumentContentUrl.newBuilder(pair._2).setId(pair._2.getContentChecksum()).build()` be more future-proof?

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/checksum/preprocessing/ChecksumPreprocessingJob.java
```diff
@@ -0,0 +1,81 @@
...
+            // Filter out records without a contentChecksum
+            JavaRDD<DocumentContentUrl> withChecksum = input.filter(
+                    doc -> doc.getContentChecksum() != null);
+
+            // Group by contentChecksum, keep record with smallest id (stable deduplication)
+            JavaPairRDD<String, DocumentContentUrl> byChecksum = withChecksum.mapToPair(
+                    doc -> new Tuple2<>(doc.getContentChecksum().toString(), doc));
+
+            JavaPairRDD<String, DocumentContentUrl> dedupedPairs = byChecksum
+                    .reduceByKey((a, b) -> a.getId().toString().compareTo(b.getId().toString()) <= 0 ? a : b);
```
mpol: The keys are not used after the deduplication, so maybe use `.values()` to just get a `JavaRDD<DocumentContentUrl>`?

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadatamerger/MetadataMergerJob.java
```diff
@@ -0,0 +1,208 @@
...
+            for (CharSequence item : second) {
+                if (!merged.contains(item)) {
+                    merged.add(item);
+                }
+            }
+        }
+        return merged.isEmpty() ? null : merged;
+    }
+
+    private static Map<CharSequence, CharSequence> mergeStringMaps(
```
mpol, line 167: It may be working for us but per https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/lang/CharSequence.html
> This interface does not refine the general contracts of the [equals](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/lang/Object.html#equals(java.lang.Object)) and [hashCode](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/lang/Object.html#hashCode()) methods. The result of testing two objects that implement CharSequence for equality is therefore, in general, undefined. Each object may be implemented by a different class, and there is no guarantee that each class will be capable of testing its instances for equality with those of the other. It is therefore inappropriate to use arbitrary CharSequence instances as elements in a set or as keys in a map.

mpol (approved):
I have a few detailed comments that could be addressed, but nothing of very high importance.

There are also two followup areas that suggest themselves:
* Some code uses `JavaRDD`, some uses `Dataset` - it would be better to chose one, preferably `Dataset`.
* Probably some jobs could be merged into fewer bigger jobs, thus writing and reading less intermediate Avro files - the generic jobs look like they would be happier as just functions called from the large jobs.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/skip_extracted/SkipExtractedJob.java
```diff
@@ -0,0 +1,98 @@
...
+            JavaRDD<ExtractedDocumentMetadata> filteredMeta = documentMeta.filter(m -> {
+                CharSequence pubType = m.getPublicationTypeName();
+                return pubType == null || !"$EMPTY$".equals(pubType.toString());
+            });
+            JavaPairRDD<String, ExtractedDocumentMetadata> filteredMetaById = filteredMeta.mapToPair(
+                    m -> new Tuple2<>(m.getId().toString(), m));
+            JavaRDD<ExtractedDocumentMetadata> outputMeta = filteredMetaById
+                    .join(contentById)
+                    .map(pair -> pair._2._1);
+
```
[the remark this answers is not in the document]
marekhorst: Good catch, the updated version of the code relies on distinct meta ids but also content ids to be fully aligned with the logic defined in the original pig script.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/skip_extracted/SkipExtractedJob.java
```diff
@@ -0,0 +1,98 @@
...
+                    params.inputDocumentMeta, ExtractedDocumentMetadata.class);
+
+            JavaPairRDD<String, DocumentContentUrl> contentById = documentContent.mapToPair(
+                    d -> new Tuple2<>(d.getId().toString(), d));
+            JavaPairRDD<String, ExtractedDocumentMetadata> metaById = documentMeta.mapToPair(
+                    m -> new Tuple2<>(m.getId().toString(), m));
+
+            // output_content: content records with no matching meta record
+            JavaRDD<DocumentContentUrl> outputContent = contentById
+                    .leftOuterJoin(metaById)
```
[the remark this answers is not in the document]
marekhorst: Good catch, the updated version of the code relies on distinct meta ids but also content ids to be fully aligned with the logic defined in the original pig script.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/skip_extracted/SkipExtractedJob.java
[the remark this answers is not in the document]
marekhorst: Btw, the same patch was applied on `SkipExtractedWithoutMetaJob` which was missing the same reduction against the metadata records.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/skip_extracted_without_meta/SkipExtractedWithoutMetaJob.java
```diff
@@ -0,0 +1,80 @@
+package eu.dnetlib.iis.wf.transformers.metadataextraction.skip_extracted_without_meta;
+
+import java.io.IOException;
+import java.util.Optional;
```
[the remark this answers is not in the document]
marekhorst: Fixed.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/idreplacer/IdReplacerJob.java
```diff
@@ -0,0 +1,104 @@
...
+import eu.dnetlib.iis.common.spark.avro.AvroDataFrameReader;
+import eu.dnetlib.iis.common.spark.avro.AvroDataFrameWriter;
+
+import static org.apache.spark.sql.functions.coalesce;
+import static org.apache.spark.sql.functions.col;
+
+/**
+ * Generic Spark job that replaces one or two id fields in an Avro dataset using an
+ * IdentifierMapping lookup. Records with no mapping entry retain a null id value
+ * in the replaced field (left-join semantics, matching PIG OUTER JOIN behaviour).
```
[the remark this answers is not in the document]
marekhorst: Fixed.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/checksum/postprocessing/meta/ChecksumMetaPostprocessingJob.java
```diff
@@ -0,0 +1,85 @@
...
+import pl.edu.icm.sparkutils.avro.SparkAvroLoader;
+import pl.edu.icm.sparkutils.avro.SparkAvroSaver;
+import scala.Tuple2;
+
+/**
+ * Spark job that remaps the id in ExtractedDocumentMetadata records from a contentChecksum
+ * back to the original document id. Only DocumentContentUrl records with a non-null
+ * contentChecksum are considered. After checksum preprocessing, the metadata id equals the
+ * checksum, so the join is metadata.id = dcu.contentChecksum → replace metadata.id with dcu.id.
+ *
```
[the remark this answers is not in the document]
marekhorst: Apparently Copilot got confused with the id->checksum mappings. The description seems to be OK to me.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/common/citations/from/referencemetadata/CitationsFromReferenceMetadataTransformerJob.java
```diff
@@ -0,0 +1,95 @@
...
+                    if (basicMetadata != null && basicMetadata.getExternalIds() != null) {
+                        externalIds = basicMetadata.getExternalIds();
+                    }
+                    CitationEntry entry = CitationEntry.newBuilder()
+                            .setPosition(ref.getPosition() != null ? ref.getPosition() : 0)
+                            .setRawText(ref.getText())
+                            .setDestinationDocumentId(null)
+                            .setConfidenceLevel(null)
+                            .setExternalDestinationDocumentIds(externalIds)
+                            .build();
```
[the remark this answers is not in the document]
marekhorst: Again, good catch.

After inspecting the code it is clear PIG script did not properly handle `null` `position` field, which is allowed by `ReferenceMetadata` schema. It didn't fail because the `position` field was never set to `null` by the code responsible for building this record.

Besides that, setting position to `0` in `CitationEntry` when `ReferenceMetadata#position` is `null` is wrong also because according to `ReferenceMetadata#position` field description in the avdl file:

> reference position in bibliography, starting with '1'

so `0` is kind of an "index out of bound" in that case.

I think the best way to solve it is to:
* fix the `ReferenceMetadata#position` field definition to disallow `null` value (which in fact is never set)
* drop checking against the `null` value in `CitationsFromReferenceMetadataTransformerJob` code discussed here.

marekhorst:
> I have a few detailed comments that could be addressed, but nothing of very high importance.
> 
> There are also two followup areas that suggest themselves:
> 
> * Some code uses `JavaRDD`, some uses `Dataset` - it would be better to chose one, preferably `Dataset`.

OK, #1600 is to meant to cover that.

> * Probably some jobs could be merged into fewer bigger jobs, thus writing and reading less intermediate Avro files - the generic jobs look like they would be happier as just functions called from the large jobs.

As discussed earlier today this is aligned with the grand plan of reducing the number of persisting operations to bare minimum. I already created a dedicated follow-up ticket for that: #1599.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/skip_extracted/SkipExtractedJob.java
marekhorst: Yes, I think at this moment, when validating the code, it could be useful but in the end PIG scripts will be removed and it hardly makes any sense to reference them. I will drop it.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/skip_extracted/SkipExtractedJob.java
marekhorst: > First, the implementation checks for "$EMPTY$" not just "EMPTY". 

Yes, this was the main issue I had to address with the first version of the generated code. Claude simply replaced `$EMPTY$` (defined in PIG code) with `EMPTY` and didn't even blink once. After discovering this issue it got fixed by Claude but it did not fix all the comments referencing this value (fixed just one below).

> Second, while the comment is technically true with the 'or' clause (as in: matches the current implementation) it is confusing as to the intent. If something is null it's definitely not "$EMPTY$", but it's a bit weird that they are opposite and not equivalent here.

I dropped "null or " from this sentence. It was kind of technically correct (in avro `CharSequence` is implemented not with `String` but `Utf8` object so checking the `null` before calling `toString()` is a good way of handling the check) but it is weird to read.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/idreplacer/IdReplacerJob.java
marekhorst: Updated.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/idreplacer/IdReplacerJob.java
marekhorst: `IdReplacerUDF` reference dropped, no need to mention it here.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/checksum/postprocessing/meta/ChecksumMetaPostprocessingJob.java
marekhorst: Dropped, `Build a mapping: checksum -> original document id` as a comment should be enough.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/checksum/preprocessing/ChecksumPreprocessingJob.java
marekhorst: Fixed, now we don't need to remember about updating it whenever schema changes.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadataextraction/checksum/preprocessing/ChecksumPreprocessingJob.java
marekhorst: Fixed.

iis-wf/iis-wf-transformers/src/main/java/eu/dnetlib/iis/wf/transformers/metadatamerger/MetadataMergerJob.java
marekhorst: Probably we were fine because objects implementing `CharSequence` handled the comparison against themselves in a proper way.

Since this is not guaranteed it may stop working once we change the environment so I have addressed it basically by trusting `String` instances more than `CharSequence` so introducing `toString()` overhead.

mpol (approved):
Looks good

marekhorst:
This feature was successfully integrated with the [k8s branch](https://github.com/openaire/iis/tree/marekhorst_1588_make_the_IIS_compliant_with_spark4_and_kubernetes_environment_APPROVED).

PR #1607, opened 2026-06-19
Closes #1606: Make affiliation matching algorithm compliant with spark4 and kubernetes environment

### LINKED ISSUE #1606: Make affiliation matching algorithm compliant with spark4 and kubernetes environment
We need to introduce support for building fat uber jar for the `iis-wf-affmatching` module by relying on the `maven-shade-plugin`.

Dedicated workflow in the https://github.com/openaire/iis-workflows repository needs to be created similarly to the `iis-wf-citation-matching-direct` module.

marekhorst:
Introducing support for building fat uber jar for the `iis-wf-affmatching` module by relying on the `maven-shade-plugin`.

This is similar change to the one previously introduced for the `iis-wf-citationmatching-direct`.

### DIFF
--- a/iis-wf/iis-wf-affmatching/pom.xml
+++ b/iis-wf/iis-wf-affmatching/pom.xml
@@ -53,6 +53,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.apache.spark</groupId>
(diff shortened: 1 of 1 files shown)


marekhorst:
The airflow workflow run on the subset of input data proved the job works as expected and it produced valid outcome.

Issue #1586, opened 2026-02-06
Integrate lens.org patents mining


marekhorst:
The following PR should be merged:

https://github.com/openaire/iis/pull/1584

integration tests suite should be revised and an appropriate case should be added.

PR #1613, opened 2026-07-01
Marekhorst 1586 integrate lens.org patents mining

marekhorst:
Bunch of details which may help in the reviewing process:
* in principle we are replacing an extremely complex, EPO-patents related, workflow involving importing patents from EPO, text mining (madis) supplemented with pantents entities generation, custom patent-related export procedure involving exporting of both document-patent relations and patent entities with the simplified version relying on predetermined set of patents coming from lens.org. Those patents are stored in an already constructed sqlite database (thus no need to import patents from EPO) and are used for text mining. There is no need to export patent entities because they will be introduced into the graph by relying on a specific workflow which is part of the openaire data provisioning system, so it happens outside of the IIS. IIS exporter module now focuses on relations only therefore it could be simplified and aligned with an already existing model (extending `AbstractActionBuilderFactory`) for exporting simple relations.
* there are changes introduced around the madis script (`patents.sql`) and the core madis (`text.py`) which are bound to the text mining logic changes after EPO patents were replaced with patents provided by lens.org
* `DocumentToPatent#lensId` is patent id provided by lens.org and it gets translated to the OpenAIRE id by the exporter module (`DocumentToPatentActionBuilderModuleFactory`)
* `PatentReferenceExtractionInput` was modified and now it represents an avro schema with raw patents metadata fully compliant with sqlitedb patents table model. Since the dynamic patent import from the graph is going to be introduced in the future (now we simply rely on a predefined sqlitedb which resides on HDFS), which means there is no `sqlite_builder` script responsible for building sqlitedb dynamically, we had to introduce a way of creating sqlitedb just for the sake of running integration tests (to be able to create a small db on the fly instead of relying on a resource which needs to be pre-uploaded on HDFS). So the current version of the `sqlite_builder` script was prepared just to cover that requirement and its content is extremelly simplified (thus `PatentReferenceExtractionInput` mimics the internal sqlitedb model). `sqlite_builder` is used in the integration tests only, it relies on a small sqlitedb (as a kind of a template for building the final database with patents) at input including `jurisdiction` predefined table.
* just realized in some commits there is #1596 referenced instead of #1586 - it needs to be fixed when sqashing


marekhorst:
Note to myself: we need to postpone the integration with the IIS master branch until having a green light from Claudio stating the remaining part of the provisioning pipeline is compliant with this change (both patents inclusion prior to IIS run and patents filtering after the mining is concluded are already in place).

iis-schemas/src/main/avro/eu/dnetlib/iis/referenceextraction/patent/PatentReferenceExtractionInput.avdl
```diff
@@ -2,6 +2,10 @@
 protocol IIS{
 
     record PatentReferenceExtractionInput {
-        string appln_nr;
+        string c1;
+        string c2;
+        union{null, string} normal = null;
+        union{null, string} c4 = null;
+        union{null, string} c5 = null;
```
mpol, line 9: Are these the best possible names for these fields?

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/DocumentToPatentActionBuilderModuleFactory.java
```diff
@@ -0,0 +1,81 @@
...
+         */
+        private AtomicAction<Relation> createAction(String source, String target, float confidenceLevel,
+                String relClass) throws TrustLevelThresholdExceededException {
+            return createAtomicActionWithRelation(source, target, OafConstants.REL_TYPE_RESULT_RESULT,
+                    OafConstants.SUBREL_TYPE_RELATIONSHIP, relClass, confidenceLevel);
+        }
+
+        private String generatePatentId(String lensId) {
+            return InfoSpaceConstants.ROW_PREFIX_RESULT + InfoSpaceConstants.LENS_ENTITY_ID_PREFIX + 
+                InfoSpaceConstants.ID_NAMESPACE_SEPARATOR +  IdentifierFactory.md5(lensId);
```
mpol, line 78: In other places `DigestUtils.md5Hex` is used instead of `IdentifierFactory.md5` - which also suggests that a functions generating ids from a prefix, ns and value could be extracted to some common utilities.

iis-wf/iis-wf-referenceextraction/src/main/java/eu/dnetlib/iis/wf/referenceextraction/patent/PatentDBBuilder.java
```diff
@@ -36,6 +38,7 @@ public ProcessExecutionContext initializeProcess(Map<String, String> parameters)
 
         String targetDbLocation = System.getProperty("java.io.tmpdir") + File.separatorChar + "patents.db";
         File targetDbFile = new File(targetDbLocation);
+        FileUtils.copyFile(new File("scripts/base_lens.db"), targetDbFile);
```
mpol, line 41: I see that there are already quite a few binary sqlite files in the repository, so maybe this is ok. But I wonder if it wouldn't be preferable to have the base data as human readable in the repository and put it in a database during runtime.

mpol (approved):
Looks good. A few things to consider as followups.

iis-wf/iis-wf-referenceextraction/src/main/resources/eu/dnetlib/iis/wf/referenceextraction/patent/main/oozie_app/workflow.xml
```diff
@@ -7,17 +7,14 @@
             <description>input document text with eu.dnetlib.iis.metadataextraction.schemas.DocumentText records</description>
         </property>
         <property>
-            <name>input_patent</name>
-            <description>input patent with eu.dnetlib.iis.referenceextraction.patent.schemas.ImportedPatent records</description>
+            <name>input_patent_db</name>
+            <value>com.cloudera.spark.lineage.NavigatorAppListener</value>
+            <description>pre-built sqlite database with patents</description>
         </property>
```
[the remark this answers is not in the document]
marekhorst: Fixed. Clearly a copy&paste issue.

iis-3rdparty-madis/src/main/resources/eu/dnetlib/iis/3rdparty/scripts/madis/functions/row/text.py
```diff
@@ -213,6 +213,17 @@ def cqlfields(*args):
 
 cqlfields.registered=True
 
+def string_split_value(s, a):
+    """
+    Returns the substring of `s` starting from the last occurrence of `a` till the end.
+    If `a` is not found in `s`, returns an empty string.
+    """
```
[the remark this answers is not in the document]
marekhorst: Updated as suggested.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/DocumentToPatentActionBuilderModuleFactory.java
```diff
@@ -0,0 +1,81 @@
...
+        // ------------------------ PRIVATE --------------------------
+        
+        /**
+         * Creates similarity related actions.
+         */
+        private AtomicAction<Relation> createAction(String source, String target, float confidenceLevel,
+                String relClass) throws TrustLevelThresholdExceededException {
+            return createAtomicActionWithRelation(source, target, OafConstants.REL_TYPE_RESULT_RESULT,
+                    OafConstants.SUBREL_TYPE_RELATIONSHIP, relClass, confidenceLevel);
+        }
```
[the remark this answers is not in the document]
marekhorst: Updated as suggested.

iis-schemas/src/main/avro/eu/dnetlib/iis/referenceextraction/patent/PatentReferenceExtractionInput.avdl
marekhorst: No, but this reflects the internal sqlite database model created by Yannis I didn't want to alter. Also because I don't really know what those names really mean which kind of confirms the names are not the best possible.

This is also not that crucial for now because currently, in production, the database is pre-built and stored on HDFS. Once Yannis prepares the `sqlite_builder` phase to build patents database on the fly from the graph this schema will be altered.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/module/DocumentToPatentActionBuilderModuleFactory.java
marekhorst: Agree. 

I'd prefer to stick to `IdentifierFactory#md5()` in all the places where md5 is calculated for OpenAIRE ids generation purposes (this utility method body could probably be simplified by simply relying on `DigestUtils.md5Hex(id)`) so I'd probably need to move it to a more generic package (`iis-commons`, because OAId generation is done in both importer and exporter sub modules) or even rely on a DNet dependency this whole `ProjectDetailConverter` seem to originate from.

Since this requires more attention, and we want to release patents mining support ASAP, I created a dedicated ticket for that: #1619.

iis-wf/iis-wf-referenceextraction/src/main/java/eu/dnetlib/iis/wf/referenceextraction/patent/PatentDBBuilder.java
marekhorst: Yes, the idea is good, I am not sure if this is easily achievable for all the cases so I'd prefer to cover it with a dedicated ticket: #1620.

marekhorst:
This feature was merged with the IIS master branch with this squashed and rebased commit: https://github.com/openaire/iis/commit/5c9cab986b4d10a2fc26f7a14883ae65388dee50.

PR #1625, opened 2026-07-28
Marekhorst 1622 make the first sqlitedb builder module runnable in the k8s environment

marekhorst:
Introducing changes in the `AbstractDBBuilder` model in order to make it more generic and easy to use in a generic airflow workflow definition by:
* fixing the names of input and output ports so they could be reused regardless of the real builder implementation
* introducing `initDbLocation` input parameter with classpath input database location whenever database needs to be updated rather than created from scratch (it is the case for patents and projects)
* moving the `initializeProcess()` final method implementation to `AbstractDBBuilder` after it got generic enough

Removing `iis-3rdparty-madis` from the project after it is expected to be baked into the docker image and available in the environment.
    
Changing madis script location in all the sqlite builder modules to `/opt/madis/`. Updating madis location in the tara reference extraction script.


marekhorst:
Merged with the https://github.com/openaire/iis/tree/marekhorst_1588_make_the_IIS_compliant_with_spark4_and_kubernetes_environment_APPROVED branch with this squashed commit: https://github.com/openaire/iis/commit/2e33e14d8461c8ceb7f5814f474420033912455a.

PR #1630, opened 2026-09-16
Marekhorst 1627 make the iis exporter compliant with spark4 and kubernetes environment

marekhorst:
This pull request introduces all the changes required to be applied on the IIS exporter codebase in order to make it executable in the k8s environment.

This covers:
* introduction of `shade-uber-jar` profiles to build uber packages for the required modules in order to reference them in airlfow workflows
* optimizing the `CitationRelationExporterJob` caching due to changes in the way cached data is serialized in spark4
* introducing the new `SequenceFileExporterJob` as a spark equivalent of the old map reduce `SequenceFileExporterMapper` which cannot be run in the k8s environment anymore
* implementing `SparkSequenceFileTestConsumer` to verify sequence files in tests

The new airflow workflows replacing the old oozie worklfows were placed in a dedicated `iis-workflows` repository under:
* https://github.com/openaire/iis-workflows/blob/main/sequencefile-exporter-on-k8s/spark_sequencefile_exporter_dag.py
* https://github.com/openaire/iis-workflows/blob/main/primary/export_dag.py

The first one is an equivalent of the old [`export/actionmanager/sequencefile`](https://github.com/openaire/iis/blob/master/iis-wf/iis-wf-export-actionmanager/src/main/resources/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/oozie_app/workflow.xml) oozie workflow while the second one binds all the exporters together and is an airflow representation of the old [`primary/export `](https://github.com/openaire/iis/blob/master/iis-wf/iis-wf-primary/src/main/resources/eu/dnetlib/iis/wf/primary/export/oozie_app/workflow.xml) oozie workflow.

### DIFF
--- a/iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/SequenceFileExporterJob.java
+++ b/iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/SequenceFileExporterJob.java
@@ -1,0 +1,171 @@
+package eu.dnetlib.iis.wf.export.actionmanager.sequencefile;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+import org.apache.avro.specific.SpecificRecordBase;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.io.Text;
+import org.apache.spark.SparkConf;
+import org.apache.spark.api.java.JavaPairRDD;
+import org.apache.spark.api.java.JavaSparkContext;
+
+import com.beust.jcommander.DynamicParameter;
+import com.beust.jcommander.JCommander;
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import eu.dnetlib.dhp.schema.action.AtomicAction;
+import eu.dnetlib.dhp.schema.oaf.Oaf;
+import eu.dnetlib.iis.common.WorkflowRuntimeParameters;
+import eu.dnetlib.iis.common.java.io.HdfsUtils;
+import eu.dnetlib.iis.common.spark.JavaSparkContextFactory;
+import eu.dnetlib.iis.common.utils.RDDUtils;
+import eu.dnetlib.iis.wf.export.actionmanager.AtomicActionSerializationUtils;
+import eu.dnetlib.iis.wf.export.actionmanager.module.ActionBuilderFactory;
+import eu.dnetlib.iis.wf.export.actionmanager.module.ActionBuilderModule;
+import eu.dnetlib.iis.wf.export.actionmanager.module.TrustLevelThresholdExceededException;
+import pl.edu.icm.sparkutils.avro.SparkAvroLoader;
+import scala.Tuple2;
+
+/**
+ * Spark 4 replacement for the MapReduce-based {@link SequenceFileExporterMapper}.
+ *
+ * <p>Reads Avro input records, converts them to {@link AtomicAction} objects via a configurable
+ * {@link ActionBuilderFactory}, and writes the JSON-serialised actions as a SequenceFile with
+ * BLOCK-level compression, preserving the output contract expected by downstream consumers.
+ *
+ * <p>The job is intended to be invoked once per export action type. The Airflow DAG submits
+ * multiple instances in parallel, one for each configured inference type, mirroring the fork/join
+ * structure of the original Oozie workflow.
+ *
+ * @author mhorst
+ */
+public class SequenceFileExporterJob {
+
+    // ----------------------- LOGIC --------------------------------
+
+    @SuppressWarnings("unchecked")
+    public static void main(String[] args) throws Exception {
+        JobParameters params = new JobParameters();
+        new JCommander(params).parse(args);
+
+        if (WorkflowRuntimeParameters.UNDEFINED_NONEMPTY_VALUE.equals(params.inputPath)) {
+            // input path is not defined for this run; nothing to export
+            return;
+        }
+
+        try (JavaSparkContext sc = JavaSparkContextFactory.withConfAndKryo(new SparkConf())) {
+            HdfsUtils.remove(sc.hadoopConfiguration(), params.outputPath);
+
+            // Capture serialisable snapshots of the driver-side state for use on executors.
+            final Map<String, String> confMap = new HashMap<>(params.dynamicParams);
+            final String factoryClassName = params.actionBuilderFactoryClassName;
+            final String schemaClassName = params.inputAvroSchemaClass;
+
+            // Resolve the concrete Avro record class on the driver (fast-fail on typos).
+            Class<? extends SpecificRecordBase> avroClass =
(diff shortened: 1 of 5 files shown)


marekhorst:
> * When `inputPath` is `$UNDEFINED$`, this return bypasses the `HdfsUtils.remove` below. The replacement DAG invokes these tasks even for undefined inputs, so rerunning with an input disabled leaves any previous SequenceFile at `outputPath`, allowing downstream export/distcp to consume stale data. Clear this path (or the run output root) before skipping, as the old workflow's prepare step did.
> 
> ```
>         if (WorkflowRuntimeParameters.UNDEFINED_NONEMPTY_VALUE.equals(params.inputPath)) {
>             // input path is not defined for this run; nothing to export
>             return;
>         }
> ```

I think we can leave it as is.

`$UNDEFINED$` indicates given inference output (one of many) is not intended to be processed. This means the output location most probably will also not be provided so there would be no valid output dir to be removed.

Besides, even if the client code provides output location while still not providing input, we could safely restrain from removing that directory because it is always randomly generated location (bound to an action set) for every provisioning round so there is no risk of including the stale data among the exporter outputs.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/SequenceFileExporterJobTest.java
```diff
@@ -0,0 +1,196 @@
...
+                .setAppName(getClass().getName())
+                .setMainClass(SequenceFileExporterJob.class)
+                .addArg("-inputPath",                     in)
+                .addArg("-outputPath",                    out)
+                .addArg("-actionBuilderFactoryClassName", factory)
+                .addArg("-inputAvroSchemaClass",          schema)
+                .addArg("-D" + paramKey,                  paramValue)
+                .addJobProperty("spark.driver.host",      "localhost")
+                .build();
+    }
```
mpol: The repetition doesn't look good but maybe extracting common code is not worth the trouble here.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/SparkSequenceFileTestConsumer.java
```diff
@@ -0,0 +1,65 @@
...
+ * <p>Reads the SequenceFile values (JSON-serialised {@link eu.dnetlib.dhp.schema.action.AtomicAction}
+ * objects) using the Spark context, counts the actual records, and asserts the count equals the
+ * expected value supplied via {@code -expectedCount}.
+ *
+ * <p>This class is intended to be run in Airflow/Kubernetes test DAGs as the validation step
+ * following a {@link SequenceFileExporterJob} invocation.
+ *
+ * @author mhorst
+ */
+public class SparkSequenceFileTestConsumer {
```
mpol: Would placing this under test sources complicate tests too much?

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/SequenceFileExporterJob.java
```diff
@@ -0,0 +1,171 @@
...
+                        // Use a lazy stream so pairs are produced on-demand rather than
+                        // accumulated in an ArrayList before being handed to the shuffle writer.
+                        // This prevents holding an entire partition's output in heap simultaneously,
+                        // which caused OOM in Celeborn's PushTask allocations on large partitions.
+                        return StreamSupport
+                                .stream(Spliterators.spliteratorUnknownSize(records, Spliterator.ORDERED), false)
+                                .flatMap(datum -> {
+                                    try {
+                                        return module.build(datum).stream()
+                                                .map(action -> {
```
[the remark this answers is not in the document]
marekhorst: This is quite relevant. It was patched in a follow-up commit.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/SequenceFileExporterJob.java
```diff
@@ -0,0 +1,171 @@
...
+            final String schemaClassName = params.inputAvroSchemaClass;
+
+            // Resolve the concrete Avro record class on the driver (fast-fail on typos).
+            Class<? extends SpecificRecordBase> avroClass =
+                    (Class<? extends SpecificRecordBase>) Class.forName(schemaClassName);
+
+            // Read all records from the Avro datastore.
+            JavaPairRDD<Text, Text> actionPairs = new SparkAvroLoader()
+                    .loadJavaRDD(sc, params.inputPath, avroClass)
+                    .mapPartitionsToPair(records -> {
```
mpol: I think I understand why mapPartitionsToPair is used here and not just mapToPair, but an additional explanatory comment would be useful.

mpol (approved):
Looks good. Some small comments only.

iis-wf/iis-wf-export-actionmanager/src/test/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/SequenceFileExporterJobTest.java
marekhorst: Yes, I agree with both remarks. Since this is just a test code I will leave it as is.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/SparkSequenceFileTestConsumer.java
marekhorst: Yes, unfortunately...

It was originally placed under the `src/test/java` but I went into a cycle of different trial and error loops to make it possible to build an uber jar for tests package but the only solution which worked was moving it under the `src/main/java`. AFAIR having to depend on test classes brought too many unwanted and problematic test dependencies. This class alone is pretty slim and I wouldn't go for it if was about to bring any test-related dependencies.

iis-wf/iis-wf-export-actionmanager/src/main/java/eu/dnetlib/iis/wf/export/actionmanager/sequencefile/SequenceFileExporterJob.java
marekhorst: It is mostly to shift from record towards the partition scale in order handle lazy streaming of records (also because there might not be any pair returned in some cases) and to initialize each executor task without going into the serialization requirement (some classes are not serializable).

marekhorst: I will add a brief version of  that remark in the comment above the relevant line.

marekhorst:
Merged with the https://github.com/openaire/iis/tree/marekhorst_1588_make_the_IIS_compliant_with_spark4_and_kubernetes_environment_APPROVED branch with the 7f5ebc85 squashed commit.

